PR #14, opened 2020-11-10
Move updates into monorepo

lboeman:


lboeman:
@alorenzo175 This is in a decent spot where the page functions as expected testing manually and the objects in `src/types/*` are tested decently. Okay with you to merge?

alorenzo175 (approved):

PR #21, opened 2020-11-24
Update models and /systems endpoints

alorenzo175:
with some dummy data and storing systems in a dict


alorenzo175:
@lboeman probably the biggest difference now is SAPMTemperatureParameters and SandiaInverterParameters, otherwise models and parameters should be similar w/ the dashboard

lboeman:
> @lboeman probably the biggest difference now is SAPMTemperatureParameters and SandiaInverterParameters, otherwise models and parameters should be similar w/ the dashboard
> 
> any other endpoints we need right now for systems?

I think this will be good for now. I'll extend the validation to use all of these models and adjust dashboard to match. Thanks for taking care of this!

lboeman (approved):

PR #25, opened 2020-11-30
Start adding db tables

alorenzo175:
Including users and systems tables.

### DIFF
--- a/db/migrations/20201130190100_create_systems_table.sql
+++ b/db/migrations/20201130190100_create_systems_table.sql
@@ -1,0 +1,23 @@
+-- migrate:up
+create table systems (
+  id binary(16) not null default (uuid_to_bin(uuid(), 1)),
+  user_id binary(16) not null,
+  definition JSON not null,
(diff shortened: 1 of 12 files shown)


db/migrations/20201130190100_create_systems_table.sql
```diff
@@ -0,0 +1,144 @@
...
+  procedure get_system (auth0id varchar(32), systemid char(36))
+    comment 'Get the definition for a system'
+    reads sql data sql security definer
+  begin
+    declare binid binary(16);
+    declare allowed boolean default false;
+
+    set binid = uuid_to_bin(systemid, 1);
+    set allowed = exists(
+      select 1 from systems where id = binid and user_id = get_user_binid(auth0id));
```
lboeman, line 29: Could use `check_users_system` here. Maybe that requires a reordering of the definitions?

lboeman (approved):
Looks good. dbmate looks really rad for tracking the migrations.

PR #26, opened 2020-12-02
Dashboard structure

lboeman:
Adds a few things:
- System listing page populated from the API.
- Edit a currently existing system spec or create a new one. 
- Login via Auth0.
- Help popups for all system spec fields.
- Adds a very simple Vuex Store, to act as a global data store for the app. This is minimally utilized at the moment, and may be useful when more data is being transferred to/from the api. 
- Reorganize component structure to move routing into a single view component that presents the header/body/footer structure and controls rendered child components inside its body.


alorenzo175:
ho boy that is a lot of changes. Can you move the auth pieces into a separate PR?

PR #27, opened 2020-12-02
description edits in model.py

cwhanse:


api/solarperformanceinsight_api/models.py
```diff
@@ -268,7 +286,7 @@ class Inverter(BaseModel):
         "", title="Make & Model", description="Make and model of the inverter"
     )
     arrays: List[PVArray] = Field(
-        ..., description="PV arrays that are connected to this inverter"
+        ..., description="Names of the PV arrays that are connected to this inverter"
```
alorenzo175, line 289:
```suggestion
        ..., description="List of PV arrays that are connected to this inverter"
```

This is actually a list of PVArray objects, not just the name. A full example of what a system looks like might be helpful https://github.com/SolarPerformanceInsight/solarperformanceinsight/blob/c041ac5babbc648df5b6e951e841b8da607cda6d/api/solarperformanceinsight_api/models.py#L8-L58

api/solarperformanceinsight_api/models.py
```diff
@@ -293,13 +311,13 @@ class PVSystem(BaseModel):
...
     elevation: confloat(ge=-300) = Field(
-        ..., description="Elevation of the system in meters"
+        ..., description="Elevation of the system above sea level in meters"
     )
     albedo: confloat(ge=0) = Field(
         ..., description="Albedo of the surface around the system"
     )
     inverters: List[Inverter] = Field(
-        ..., description="Inverters that are connected to make up this system"
+        ..., description="Names of the inverters that make up this system"
```
alorenzo175, line 320:
```suggestion
        ..., description="List of inverters that make up this system"
```

alorenzo175 (commented):
We've been using [black](https://github.com/psf/black) so we don't have to think about formatting. You can install it and run it on the code to autoformat it, or use [pre-commit](https://pre-commit.com/) and `pre-commit install` to automatically run it before a git commit

alorenzo175 (approved):

PR #29, opened 2020-12-03
Update model fields

alorenzo175:
to play nicely with mypy and add a specific system example to OpenAPI. 

Will rebase on #27 once that's in

### DIFF
--- a/api/solarperformanceinsight_api/models.py
+++ b/api/solarperformanceinsight_api/models.py
@@ -382,24 +382,34 @@ class PVSystem(BaseModel):
         schema_extra = {"example": SYSTEM_EXAMPLE}
 
 
-class PVSystemID(BaseModel):
-    system_id: UUID = Field(..., description="Unique identifier of the system")
+class StoredObject(BaseModel):
+    object_id: UUID = Field(..., description="Unique identifier of the object")
+    object_type: str = Field("system", description="Type of the object")
+    created_at: dt.datetime = Field(..., description="Datetime the object was created")
+    modified_at: dt.datetime = Field(
+        ..., description="Datetime the object was last modified"
+    )
 
     class Config:
         schema_extra = {
-            "example": {"system_id": "6b61d9ac-2e89-11eb-be2a-4dc7a6bcd0d9"}
+            "example": {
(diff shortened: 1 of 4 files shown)


api/solarperformanceinsight_api/models.py
```diff
@@ -285,23 +358,50 @@ class Inverter(BaseModel):
...
+    class Config:
+        schema_extra = {
+            "example": {"system_id": "6b61d9ac-2e89-11eb-be2a-4dc7a6bcd0d9"}
+        }
+
+
+class StoredPVSystem(PVSystemID):
+    created_at: dt.datetime = Field(..., description="Datetime system was created")
+    modified_at: dt.datetime = Field(..., description="Datetime system last modified")
+    definition: PVSystem
```
alorenzo175, line 397: @lboeman This change will definitely affect the dashboard. I think it makes sense to separate the pvsystem definition from storage metadata (system_id, created_at, modified_at, maybe others in the future). I think this will make life easier when we need to process the data i.e. just pass storedpvsystem.definition vs the whole thing with the extra fields. What do you think?

lboeman: I think this is the right way to go. 
One thing that was kind of a pain point on the arbiter dashboard was different uuid field names. Particularly when trying to write generic render code for similar objects with minor tweaks. It might make things a little easier if we had a pattern like `{'type': 'system', 'id': <uuid>}` over a field name like `system_id`, which seems to follow the pattern of separation of the storage metadata and object specification.
But since we're not storing too much here that might not be such an issue and I'm not opposed to just continuing with this pattern as this is cleaner on the python side. Just something to consider going forward.

alorenzo175: > It might make things a little easier if we had a pattern like {'type': 'system', 'id': <uuid>} over a field name like system_id, which seems to follow the pattern of separation of the storage metadata and object specification.

sounds good to me

alorenzo175: actually decided to use `object_id` and `object_type` to avoid conflicting w/ the builtin python functions

lboeman (approved):

PR #30, opened 2020-12-03
Dashboard Auth setup

lboeman:
breaking apart #26 into smaller chunks, this includes auth0 setup and a Vuex store that lets us maintain a list of systems from the api. Almost all `.vue` files can be ignored when reviewing except for Home.vue which contains a little bit of authentication code. Routes are protected in by using `beforeEnter: authGuard`


dashboard/src/auth/VueAuth.ts
```diff
@@ -0,0 +1,120 @@
...
+    onRedirectCallback: RedirectCallback,
+    redirectUri: string,
+    auth0Options: Auth0Options
+  ) {
+    // Create a new instance of the SDK client using members of the given options object
+    this.auth0Client = await createAuth0Client({
+      domain: auth0Options.domain,
+      client_id: auth0Options.clientId, // eslint-disable-line @typescript-eslint/camelcase
+      audience: auth0Options.audience,
+      redirect_uri: redirectUri // eslint-disable-line @typescript-eslint/camelcase
```
alorenzo175, line 89: I think we probably want to use refresh tokens and I think auth0 is already configured for them

https://auth0.com/docs/libraries/auth0-single-page-app-sdk#use-rotating-refresh-tokens

alorenzo175 (commented):
Should we try and test any of the auth stuff?

dashboard/src/auth/VueAuth.ts
lboeman: Yes, I'll add some auth tests and add refresh token use.

alorenzo175 (approved):

PR #32, opened 2020-12-03
Restructure dashboard

lboeman:
Will need a Rebase on #31 

Changes summary:
- Reorganize components directory, with generic/helper components in the root and a `model` directory that contains all of the vue components for editing objects found in `src/types`.
- Update components to use new api fields.
- Create `Home.vue` view component that contains header, footer, nav, and a body element that renders the current route's component. Nav displays differently for authenticated/unauthenticated users.

A large portion of this PR is the unfortunate input element pattern like:
```
 <b>Make and Model:</b>
  <input v-model="pvarray.make_model" />
  <help :helpText="this.definitions.properties.make_model.description" />
  <br />
  <span style="color:#F00;" v-if="'make_model' in this.errors">
    {{ this.errors.make_model }}
    <br />
  </span>
```
I haven't found a good way to factor out this pattern that isn't just as verbose and more obscure so I've left it for now.


alorenzo175:
> A large portion of this PR is the unfortunate input element pattern like:
> 
> ```
>  <b>Make and Model:</b>
>   <input v-model="pvarray.make_model" />
>   <help :helpText="this.definitions.properties.make_model.description" />
>   <br />
>   <span style="color:#F00;" v-if="'make_model' in this.errors">
>     {{ this.errors.make_model }}
>     <br />
>   </span>
> ```
> 
> I haven't found a good way to factor out this pattern that isn't just as verbose and more obscure so I've left it for now.

Sorry I didn't really give this much thought before, but it's repeated so often that it'd be easy to miss something now or if something needs to change later. So, even if it's about as verbose, is it possible to make a component like this?
```
<parameter-field :title="Make and Model" :model="pvarray.make_model" :key="make_model"></parameter-field>
```

dashboard/src/components/model/Inverter.vue
```diff
@@ -24,30 +39,39 @@
...
 import {
-  PVSystInverterParameters,
+  SandiaInverterParameters,
   PVWattsInverterParameters
 } from "@/types/InverterParameters";
 
 Vue.component("arrays-view", ArraysView);
 Vue.component("inverter-parameters", InverterParametersView);
+Vue.component("loss-parameters", LossParametersView);
+Vue.component("help", HelpPopup);
```
alorenzo175, line 59: Should all of these components be registered globally or at a higher level instead of in each component?

https://vuejs.org/v2/guide/components-registration.html#Global-Registration

dashboard/src/components/model/LossParameters.vue
```diff
@@ -0,0 +1,135 @@
...
+      </span>
+      <b>snow:</b>
+      <input type="number" v-model.number="parameters.snow" />
+      <help :helpText="this.definitions.properties.snow.description" />
+      <br />
+      <span style="color: #F00;" v-if="'snow' in this.errors">
+        {{ this.errors.snow }}
+        <br />
+      </span>
+      <b>mismath:</b>
```
alorenzo175, line 28:
```suggestion
      <b>mismatch:</b>
```

dashboard/src/views/SystemSpec.vue
```diff
@@ -0,0 +1,218 @@
...
+
+      if (SandiaInverterParameters.isInstance(firstParams)) {
+        this.model = "pvsyst";
+      } else if (PVWattsInverterParameters.isInstance(firstParams)) {
+        this.model = "pvwatts";
+      }
+    }
+  }
+
+  async loadSystem() {
```
alorenzo175, line 147: shouldn't the system already be in the store? In any case, I think a separate interface to the API detached from views or components might be helpful. Maybe that is a future issue

dashboard/src/views/SystemSpec.vue
```diff
@@ -0,0 +1,218 @@
...
+    const token = await this.$auth.getTokenSilently();
+    const response = await fetch(`/api/systems/${this.systemId}`, {
+      headers: new Headers({
+        Authorization: `Bearer ${token}`
+      })
+    });
+    if (response.ok) {
+      const system: System = await response.json();
```
alorenzo175, line 155: this is really a StoredSystem. So for now the system should be the .defnition attribute of the returned json

dashboard/src/components/model/ModuleParameters.vue
```diff
@@ -0,0 +1,205 @@
...
+        <br />
+      </span>
+      <b>mu_gamma:</b>
+      <input v-model="parameters.mu_gamma" />
+      <help :helpText="this.definitions.properties.mu_gamma.description" />
+      <br />
+      <span style="color:#F00;" v-if="'mu_gamma' in this.errors">
+        {{ this.errors.mu_gamma }}
+        <br />
+      </span>
```
alorenzo175, line 36: Why does gamma_ref have `type="number" v-model.number=` but mu_gamma doesn't?

alorenzo175: maybe this is causing one validation issue (along with the added example in the openapi spec)

image

dashboard/src/store/actions.ts
```diff
@@ -1,15 +1,19 @@
...
     const token = await this._vm.$auth.getTokenSilently();
     const systemlist = await fetch("/api/systems/", {
       headers: new Headers({
         Authorization: `Bearer ${token}`
       })
     }).then(response => response.json());
-    context.commit("updateSystemsList", systemlist as Array<System>);
+    const systemMapping: Record<string, System> = {};
+    for (const system of systemlist) {
+      systemMapping[system.system_id] = system as System;
```
alorenzo175, line 15: api has changed so this should be 
```suggestion
      systemMapping[system.object_id] = system.definition as System;
```
or whatever for typescript etc

dashboard/src/components/ModelField.vue
```diff
@@ -0,0 +1,32 @@
...
+  //@Prop() errors!: Record<string, any>;
+  //@Prop() definitions!: Record<string, any>;
+
+}
+</script>
+
+<!-- Add "scoped" attribute to limit CSS to this component only -->
+<style scoped>
+
+</style>
```
lboeman, line 32: @alorenzo175 I created this component to factor out the copy pasta. The way it's written here, it reads and alters the props of its parents directly. We could also use the commented out props and pass all of the props in. I think this should be okay since this will always be a leaf node with no children, and I can't think of any issues it may cause. Do you have any objections or concerns with this?

alorenzo175: I'm definitely just learning Vue, so how is this altering the parent?

Also, I don't think the errors span is working. For example, modifying models.py adding `ge=0` to `u_c` and then entering -1 for uC does not generate 'should be >= 0'

lboeman: Ah, thanks. there was no validate function for TemperatureParameters. errors should work now, an empty field will yield "should be a number".

The `v-model` bindings are accessing the `$parent` directly. Essentially, anywhere this is used is agreeing to have it's `parameters` mutated by this component. It's not too far from what the rest of these components are doing, e.g. clicking "add inverter" appends a new inverter object to the systems `inverters` property, but hands it off to a component to be altered.

To explain why I'm double checking, All of the model components break from the suggested Vue pattern of passing props down through the hierarchy, and emitting events up to mutate data. But since these components are meant to mutate the system specification, which will only exist in memory on the page, it seems like a logical way to go.

alorenzo175: I see. I guess that's OK. good to go change the rest to use this component

alorenzo175:
Looks good, but on Firefox + fedora only integers are allowed in the "number" fields. See toaster photo

alorenzo175:
Apparently not a problem in midori, but is a problem in chrome with the hover showing up but no red outline

alorenzo175:
ah, may need to set the `step="0.01"` or similar https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number.

lboeman:
> ah, may need to set the `step="any"` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number.
> 
> ~It seems that you need to set the step to the smallest decimal place allowed to avoid the warning~

Thanks, added. I also realized I was missing the `strings` and `modules_per_string` field, which have a type `integer` in the spec, so I've added that.

lboeman:
This does have a bug that causes errors with similarly named fields to trigger errors upstream in the model. For example, setting an array name to "name*" triggers errors on inverter names and the system name. This is caused by a naive lookup of the field name in the errors object. But, I think there's probably a better generic way to do it that will work with API errors as well, so I'll hold off for another PR to fix that particular bug.

alorenzo175:
> This does have a bug that causes errors with similarly named fields to trigger errors upstream in the model. For example, setting an array name to "name*" triggers errors on inverter names and the system name. This is caused by a naive lookup of the field name in the errors object. But, I think there's probably a better generic way to do it that will work with API errors as well, so I'll hold off for another PR to fix that particular bug.

Ok, let's make an issue for that

alorenzo175 (approved):

PR #43, opened 2020-12-08
add a /systems/check endpoint to validate json

alorenzo175:
without saving the system. Does not check a ModelChain object yet in the interest of keeping PRs small


lboeman (approved):
Looks good!

PR #47, opened 2020-12-09
Construct ModelChain objects from models.PVSystems

alorenzo175:
We'll use the functions in `pvmodeling` later to construct and run the ModelChains. For now, I've restricted the number of arrays per inverter to 1 until pvlib 0.9. I also added:

- the pvlib models corresponding to the parameter sets as the `_pvib_*_model` private attribute, e.g. `SAPMTemperatureParameters._pvlib_temperature_model == 'sapm'`
- the ability to specify the other ModelChain models like aoi, clearsky, etc

### DIFF
--- a/api/solarperformanceinsight_api/models.py
+++ b/api/solarperformanceinsight_api/models.py
@@ -1,6 +1,7 @@
 import datetime as dt
 from typing import Union, List, Optional, Any
 from pydantic import BaseModel, Field
+from enum import Enum
 from pydantic.fields import Undefined
 from pydantic.types import UUID
 
@@ -53,6 +54,11 @@ SYSTEM_EXAMPLE = dict(
                     ),
                 )
             ],
+            airmass_model="kastenyoung1989",
+            aoi_model="no_loss",
+            clearsky_model="ineichen",
+            spectral_model="no_loss",
+            transposition_model="haydavies",
         )
     ],
 )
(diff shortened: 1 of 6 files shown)


api/solarperformanceinsight_api/models.py
```diff
@@ -168,6 +174,7 @@ class PVsystModuleParameters(BaseModel):
         ),
         gt=0,
     )
+    _pvlib_dc_model: str = PrivateAttr("pvsyst")
```
wholmgren, line 177: might want to name these attributes `_modelchain_dc_model` to avoid confusion with the pvlib function names.

wholmgren: and similar for the other `_pvlib*` attributes

api/solarperformanceinsight_api/pvmodeling.py
```diff
@@ -0,0 +1,95 @@
...
+
+def extract_pvlib_models(inverter: models.Inverter) -> Dict[str, str]:
+    """Extract the pvlib.modelchain.ModelChain _model keywords from an
+    Inverter definition"""
+    models = dict(inverter._pvlib_models)
+    models.update(
+        {
+            k: v
+            for k, v in inverter.dict().items()
+            if k.endswith("_model") and k != "make_model"
```
wholmgren, line 78:
```suggestion
            if k.startswith("_pvlib") and k.endswith("_model")
```

api/solarperformanceinsight_api/pvmodeling.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    return models
+
+
+def construct_modelchains(system: models.PVSystem) -> List[ModelChain]:
+    """Construct a pvlib.modelchain.ModelChain object for each Inverter
+    in system"""
+    location = construct_location(system=system)
+    out = []
+    for i, inverter in enumerate(system.inverters):
```
wholmgren, line 89: why the enumeration? `i` is not used.

wholmgren (approved):
looks correct. interesting to see the formalization. I've been wanting that kind of thing as I wade through the changes in the pvlib array pr.

api/solarperformanceinsight_api/pvmodeling.py
alorenzo175: right, I was getting ahead of myself there thinking about descriptive error messages (w/ location of problematic inverter) if the ModelChain construction failed. I don't think that failure is currently possible with our strict schema until maybe multiple arrays are implemented

api/solarperformanceinsight_api/pvmodeling.py
alorenzo175: ln 73 actually handles the `_pvlib` models, this is for `aoi_model` etc. It actually probably makes the most sense to just move all of that to `models.Inverter.__init__` and `models.Inverter._modelchain_models`

PR #48, opened 2020-12-09
Move the albedo parameter from a System attribute to a PVArray attribute

alorenzo175:
This is more consistent with the mapping PVArray -> pvlib.pvsystem.PVSystem (and later pvsystem.Array). I'm sure we could also write some javascript to set the albedo of a newly added array to that of the first array on the dashboard. @cwhanse @wholmgren  what do you think of this schema change?

### DIFF
--- a/api/solarperformanceinsight_api/models.py
+++ b/api/solarperformanceinsight_api/models.py
@@ -11,7 +11,6 @@ SYSTEM_EXAMPLE = dict(
     latitude=33.98,
     longitude=-115.323,
     elevation=2300,
-    albedo=0.2,
     inverters=[
         dict(
             name="Inverter 1",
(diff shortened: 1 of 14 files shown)


wholmgren:
No objection to changing it, especially if it's simple for users. Here's some copied discussion from the spec google doc about a month ago:

Cliff Hansen
Cliff Hansen
2:59 PM Nov 3
I think albedo could be system-level.
William Holmgren
William Holmgren
3:52 PM Nov 4
I put it here because it's an array attribute in the pvlib pr. I'm fine with either location for SPI wrapper.
Cliff Hansen
Cliff Hansen
9:14 AM Nov 5
I'm fine with albedo in either location, it will be unusual to see different values among arrays but doesn't hurt.

lboeman (approved):

PR #49, opened 2020-12-09
Add basic model component tests

lboeman:
Mounts and checks that all of the expected fields are represented in components so we don't leave out fields.


lboeman:
Probably need to adjust the `$parent` access in the ModelField component to test completely. Parent components are not injected correctly by vue-test-utils, so passing in the props `parameters`, `definitions`, and `errors` will make proper testing possible. This is a good start though, and could be adjusted with minimal changes.

alorenzo175 (approved):

PR #50, opened 2020-12-09
Pass props into ModelField component

lboeman:
Reverts decision to use `$parent` in model field, instead opting to pass in props from the parent component. 

Adds a copy of the `openapi.json` for mocking api call for validation.

Tests are updated to ensure that model fields exist and that the values match those of the passed in parameters.


lboeman:
Working on adding tests for changes/watchers but I think in the interest of keeping this PR as small as it can be this is at a good point for review/merge.

dashboard/src/components/ModelField.vue
```diff
@@ -12,14 +12,17 @@ ON THE SAME FIELD.
...
-property name to render as the `field-name` Prop.
+Parent components should v-bind `parameters`, `definitions`, and `errors` and
+pass the name of the property to render as the `fieldName` prop.
 
 e.g. For the PVWatts Inverter Parameter schema, the pdc0 field can be rendered
      with:
 
-    ` <model-field field-name="pdc0">`
+    <model-field
+      :parameters="paremeters"
```
alorenzo175, line 22:
```suggestion
      :parameters="parameters"
```

alorenzo175 (approved):

PR #58, opened 2020-12-10
add a script to update js assets from the latest api

alorenzo175:
right now this includes dashboard/tests/unit/openapi.json.
The extract_files.py script also has a check command to make
sure the files are up to date w/ each python workflow run


alorenzo175:
@lboeman the inverters.json is somewhat large, and I'm not sure how js will do with processing it. Another option would be to add dynamic endpoints like /inverters/keys and /inverters/{inverter name} so we wouldn't need to send much data

lboeman:
> @lboeman the inverters.json is somewhat large, and I'm not sure how js will do with processing it. Another option would be to add dynamic endpoints like /inverters/keys and /inverters/{inverter name} so we wouldn't need to send much data

I think that sounds good. Maybe with the name of the schema added to the path so if we have to extend it we don't have to alter the API? Something like `/inverters/SandiaInverterParameters/keys`. 

On the dashboard side, I think the inverter browser should probably be it's own component that pops up when a user hits a "browse inverter database" button and does the following:
 - Loads and displays the list of names with a search bar.
 -  When a user selects an inverter, it fetches the parameters from the api and displays them.
 - User hits "use these parameters" to load those values into the form, or can select another inverter/ cancel and use their own.

alorenzo175:
> I think that sounds good. Maybe with the name of the schema added to the path so if we have to extend it we don't have to alter the API? Something like /inverters/SandiaInverterParameters/keys.

good idea

> On the dashboard side, I think the inverter browser should probably be it's own component that pops up when a user hits a "browse inverter database" button and does the following:
>
>    Loads and displays the list of names with a search bar.
>    When a user selects an inverter, it fetches the parameters from the api and displays them.
>    User hits "use these parameters" to load those values into the form, or can select another inverter/ cancel and use their own.

That all sounds good

lboeman (approved):

PR #59, opened 2020-12-11
add /parameters/inverters/sandia/ endpoints

### LINKED ISSUE #39: extract pvlib inverter and module parameters
into a format like JSON to be consumed by the dashboard to assist in defining systems

alorenzo175:
to get available inverters and their parameters rather than relying
on a large json file. closes #39


api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -0,0 +1,49 @@
...
+
+from .. import models
+
+
+router = APIRouter()
+sandia_inverter_params = (
+    retrieve_sam("CECInverter")
+    .loc[models.SandiaInverterParameters.schema()["properties"].keys()]
+    .astype(float)
+)
```
lboeman, line 18: Can this be a dict or function that is keyed by the name of the model in `models.py` and returns the set of appropriate parameters?
e.g. `{ "SandiaInverterParameters" : retrieve_sam(... }`. 

That would make it easier to access from the dashboard. And users would be able access the parameters using the name of the component described by the api docs.

If this needs to be something the api needs to be able to load based on the inverter nameand the private `_modelchain`  attrs then This looks good.

lboeman: Then we could use routes like `/parameters/{schema name}/{inverter name}`. And we'd be able to have one endpoint for multiple schemas.

lboeman: Yes, sorry, that is what I was trying to get at.

alorenzo175: ok, the routes are now `/parameters/sandiainverterparameters` and `/parameters/sandiainverterparameters/{inverter_name}`. Keeping the camelCase for this and lowercase for the rest bugged me. Also defined these routes/functions explicitly instead of `/parameters/{schema_name}/{name}` for easier to understand API docs

lboeman (approved):

Issue #46, opened 2020-12-08
either update system name along with a definition or restrict the name field


alorenzo175:
and don't allow updates to the name. Right now, one can modify the name in an update request but the name is not modified in the database `name` column which could cause conflicts later since (name, user_id) is a unique index

lboeman:
closed in #61

PR #61, opened 2020-12-11
Fix system loading

lboeman:
Load systems from the api on page load to make sure the `/system/uuid` endpoint can be hit by knowing the url. The Vuex store is only populated when `/systems` is visited. 

Displays the system name when editing an existing system to avoid conflict response mentioned in #46 

Adds a section that prints apiErrors on post failures. Currently just dumps the response JSON for easier debugging. 

Adds system not found error if a 404 is returned, and doesn't render the form.

### DIFF
--- a/dashboard/src/components/model/System.vue
+++ b/dashboard/src/components/model/System.vue
@@ -1,6 +1,11 @@
 <template>
   <div class="system">
+    <div v-if="exists" class="model-field">
+      <b>Name:</b>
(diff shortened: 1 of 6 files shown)


alorenzo175 (commented):
This is great! can you add tests for saveSystem?

lboeman:
Tests Added!

alorenzo175 (approved):
one more test for updating a system

PR #66, opened 2020-12-14
Inverter browser

lboeman:
Adds an inverter database browser popup that show up when the user clicks a "browse inverter database button". The list of inverter names is shown and searchable, however search requires the user to click "search" or hit enter instead of reacting to their typing. Due to the number of objects available in this and future databases, updating reactively would be an issue. 
Once the user selects an inverter, they can preview the parameters and click "use these parameters" to import them into the form.
Screenshot from 2020-12-14 11-56-31
Screenshot from 2020-12-14 11-56-44

### DIFF
--- a/dashboard/src/components/Browser.vue
+++ b/dashboard/src/components/Browser.vue
@@ -76,13 +76,9 @@ export default class DBBrowser extends Vue {
     });
   }
   async loadOptions() {
-    const token = await this.$auth.getTokenSilently();
     const response = await fetch(
-      `/api/parameters/${this.componentName.toLowerCase()}`, {
-        headers: new Headers({
-          Authorization: `Bearer ${token}`
-        })
-    });
+      `/api/parameters/${this.componentName.toLowerCase()}`,
+    );
     const optionList = await response.json()
     this.options = optionList;
     this.selectOptions = optionList;
@@ -93,13 +89,9 @@ export default class DBBrowser extends Vue {
     this.filterOptions();
   }
   async fetchSpec(){
-    const token = await this.$auth.getTokenSilently();
     const response = await fetch(
-      `/api/parameters/${this.componentName.toLowerCase()}/${this.selection}`, {
-        headers: new Headers({
-          Authorization: `Bearer ${token}`
-        })
-    });
+      `/api/parameters/${this.componentName.toLowerCase()}/${this.selection}`
+    );
     const optionList = await response.json();
     return optionList;
   }
--- a/dashboard/src/components/Browser.vue
+++ b/dashboard/src/components/Browser.vue
@@ -47,7 +47,6 @@ export default class DBBrowser extends Vue {
   spec: Record<string, any> = {};
 
   mounted() {
-    console.log("browser mounted");
     this.loadOptions();
   }
   data() {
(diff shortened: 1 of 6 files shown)


dashboard/src/components/Browser.vue
```diff
@@ -0,0 +1,142 @@
...
+  }
+  filterOptions(){
+    this.optionsLoading = true;
+    this.setFilteredOptions().then((opts) => {
+      this.optionsLoading = false;
+      this.selectOptions = opts;
+    });
+  }
+  async loadOptions() {
+    const token = await this.$auth.getTokenSilently();
```
alorenzo175, line 79: unnecessary. endpoint doesn't require auth

dashboard/src/components/Browser.vue
```diff
@@ -0,0 +1,142 @@
...
+    this.options = optionList;
+    this.selectOptions = optionList;
+    this.optionsLoading = false;
+  }
+  resetSearch(){
+    this.search = "";
+    this.filterOptions();
+  }
+  async fetchSpec(){
+    const token = await this.$auth.getTokenSilently();
```
alorenzo175, line 96: unnecessary

dashboard/src/components/model/InverterParameters.vue
```diff
@@ -111,25 +108,19 @@ export default class InverterParametersView extends ModelBase {
...
-    // Return a list of parameter options based on currently selected model
-    if (this.model == "pvsyst") {
-      return ["PVSystInverter_0", "PVSystInverter_1"];
-    } else {
-      return ["PVWattsInverter_0", "PVWattsInverter_1"];
-    }
+  loadInverter(parameters: Record<string, any>) {
+    this.showBrowser = false;
+    // only supported for pvsyst model
+    this.parameters = new SandiaInverterParameters(parameters);
```
alorenzo175, line 123: this must not propagate correctly or something. If you add the parameters, then click Add Array, they disappear. If you add an array, then add parameters, then save, the parameters are not saved

alorenzo175: I think that failing test gets at this

alorenzo175 (commented):
I think the interface is good to get started, but should probably be tested

dashboard/src/components/model/InverterParameters.vue
lboeman: Thanks, this wasn't working correctly. Setting the whole Prop at once in inverter parameters was breaking the reference to the passed in `inverter_parameters`, so I've moved the event handling up a level, so the Inverter component can assign it's own parameters and then pass them back down for UI update. This is one of the tricky direct prop mutation issues. Okay to adjust the values that each object's properties point to but need to be careful to not destroy pointers.

lboeman:
Tests are written, and this is pretty much ready. I can't figure out a bug where clicking on an inverter name causes the cursor to start selecting any option below it until clicking out of the `<select>` element and fires an event for each selection. It's happening for Chrome and Opera but not Firefox.

dashboard/src/components/Browser.vue
```diff
@@ -0,0 +1,136 @@
...
+      selectOptions: this.options,
+      selection: this.selection,
+      spec: this.spec,
+      search: this.search
+    };
+  }
+  async setFilteredOptions() {
+    let opts: Array<string>;
+    if (this.search && this.search != "") {
+      opts = this.options.filter(x => x.includes(this.search));
```
alorenzo175, line 67: can we ignore case here and maybe make an issue for fuzzier searching?

alorenzo175 (approved):
Works great for me

Issue #65, opened 2020-12-12
interface for starting analysis workflow


wholmgren:
How are we going to present various analyses pathways to the user? A few options:

A. Select workflow, then select site, then upload data. User clicks buttons for one of **Calculate predicted performance**, **Calculate expected performance**, etc. Next page shows the table of sites with previously defined metadata, and user clicks one of these names, clicks "create and save a new site", or uploads a json file. For selection of an existing site, next page shows the relevant site layout with buttons for uploading time series data and associating data with the right objects. For selection of create a new site, go to standard site creation page, but when user clicks "save site" they go directly to the relevant time series data input page. For "uploads a json file", API validates the input and then goes directly to the time series data input page. Pros: easier for new users. ~Implies API routes like `/workflow1` that naturally accept the metadata and data in one post.~ Keeps the metadata management very distinct from the workflows. Cons: more clicking. 

B. Select site, then select workflow, then upload data. User sees the table of sites with previously defined metadata and selects one of these or clicks "create a new site". Click on existing site shows the detailed metadata, with buttons **Edit metadata**, **Calculate predicted performance**.... Same time series data input pages follow. Pros: ? Cons: site page becomes even more complicated. ~Implies API routes like `/site1/workflow1` and that's inconsistent with the idea that you'd simultaneously post metadata and data.~

C. Select workflow for a particular site, then upload data. User sees the table of sites with previously defined metadata, and there are buttons for **Calculate predicted performance**, etc. next to each name. Pros: all options available from a single page. Cons: requires that user understands the entire workflow before doing anything. ~To me, this also implies API routes like `/site1/workflow1`.~

(strike-through because @alorenzo175 explained to me that the REST API considerations of this discussion are irrelevant. new issue being created soon for `/jobs/` API explanation.)

Also worth revisiting the use cases with a focus on the time series data:

- [1.A. Calculate predicted performance](https://solarperformanceinsight.org/usecases/#uc1A). One upload: 
  - original weather data

- [1.B. Calculate expected performance](https://solarperformanceinsight.org/usecases/#uc1B). One upload: 
  - actual weather data. Data quality checks may be applied if implemented (#62).

- [1.C Account for snow and soiling](https://solarperformanceinsight.org/usecases/#uc1C). I don't think we need a special workflow for this (also these models still need to be hooked into ModelChain).

- [2.A Compare predicted and actual performance](https://solarperformanceinsight.org/usecases/#uc2A). Three uploads: 
[...]

PR #68, opened 2020-12-15
Mock up system page with workflow buttons

lboeman:
Working towards #65 
Changes system management table contain Name, Edit, Delete, Predict Performance, Compare Performance, and Calculate PR buttons. Calculation buttons aren't wired up to anything yet, I'll work on mocking up some basic calculation pages next. 
Currently looks like: 
Screenshot from 2020-12-15 12-30-23
When the user clicks "Delete" they're promped with a modal to confirm they want to actually delete the system:
Screenshot from 2020-12-15 12-30-31


lboeman:
I've dropped in some stub pages that contain a list of steps for the workflow that goes there: 
Screenshot from 2020-12-15 15-21-02

Users would click the "Predict performance" button next to a system, land on a page with something similar to the "I want to:" selection above, and based on their selection receive prompts for the data needed to continue.
@wholmgren Any thoughts before this gets any more fleshed out? 

Thinking about the jobs system and the need to possibly return to a partially finished job, I realized that these pages are really just going to serve as the job registration page. A user will click some button to start a workflow, which will register a job with the api, then hand off the job id to another Component (that could be nested on the same page and reused elsewhere) that handles the job setup, submission and report retrieval.

wholmgren:
Is the other option in the drop down menu "calculate expected performance"? Assuming yes, two comments:

1. It's weird that I'd click a button that says "predict performance" and then I'd a menu with those two phrases.
2. I don't think the average user is going to appreciate the distinction. 

It might be more coherent if we say:

```
I want to calculate the system output using...

<a>weather data provided when the system was designed</a>
<a>actual weather data during system operation</a>
```

So we're kind of combining the ideas in #65.A (sorry I can't let it go) and #65.C

Also has me wondering if the button in the table should be "Calculate performance" instead of "Predict Performance" since "calculate" is the common word in those use case descriptions.

lboeman:
> ```
> I want to calculate the system output using...
> 
> <a>weather data provided when the system was designed</a>
> <a>actual weather data during system operation</a>
> ```

That's much better, and makes the difference very clear.

> So we're kind of combining the ideas in #65.A (sorry I can't let it go) and #65.C
> 
> Also has me wondering if the button in the table should be "Calculate performance" instead of "Predict Performance" since "calculate" is the common word in those use case descriptions.

I agree. "Calculate performance" is more fitting here. Thanks!

lboeman:
Added another section to the workflow "get started" page to account for temperature options:
Screenshot from 2020-12-17 09-44-00
My understanding is that `module_temperature` or `cell_temperature` will suffice on their own or the user needs to supply `wind_speed` and `temp_air`. Here I've left off "in my data" from the air temp/wind speed option for supporting the user inputting a constant. 
@wholmgren Does this seem correct?

cwhanse:
I like the style here.

The desired result is cell temperature.

If cell_temperature is supplied, no other data is needed and no model should be run.
If module_temperature is supplied, then a model (pvlib.temperature.sapm_cell_from_module) should be used. The model needs plane-of-array irradiance data from something upstream (user upload, or modeled from user-provided ghi, dhi, etc.)
If neither cell_temperature or module_temperature is supplied, then cell_temperature should be calculated from the user's weather data (irradiance, air temperature and wind speed).

"How should we determine cell temperature?"
o Cell temperature is included in my data.
o Calculate cell temperature from module temperature and irradiance in my data.
o Calculate cell temperature from irradiance, air temperature and wind speed in my data.

wholmgren:
maybe another topic for the pvlib modelchain.rst rewrite

lboeman:
> I like the style here.
> 
> The desired result is cell temperature.
> 
> * If cell_temperature is supplied, no other data is needed and no model should be run.
> * If module_temperature is supplied, then a model (pvlib.temperature.sapm_cell_from_module) should be used. The model needs plane-of-array irradiance data from something upstream (user upload, or modeled from user-provided ghi, dhi, etc.)
> * If neither cell_temperature or module_temperature is supplied, then cell_temperature should be calculated from the user's weather data (irradiance, air temperature and wind speed).
> 
> "How should we determine cell temperature?"
> o Cell temperature is included in my data.
> o Calculate cell temperature from module temperature and irradiance in my data.
> o Calculate cell temperature from irradiance, air temperature and wind speed in my data.

Thanks for the clarification!

lboeman:
I've added a short metadata section to each of of the mapping sections, and made the mapper work for multiple inverters/arrays.

I've also added a warning to users when the number of columns in their file is less than the number of required variables times the number of arrays or inverters. This is still assuming one file containing all of the weather data.

Here's a screenshot of the mapping page for a 2 inverter, 2 arrays per inverter system with a data at the array level.
4arraymapping

lboeman:
Some style updates and tweaks to functionality:
- Converted names of variables to be mapped to something more human readable e.g. "ghi" became "Global Horizontal Irradiance".
- Changed display name of the `make_model` field in Inverter and Array objects to be "Inverter Make and Model" and "Module Make and Model" respectively. These workflow pages have those values hard-coded but the System definition spec pulls those from the api, so #72 adjusted them.
- Moved "Module Make and Model" to just above module parameters on the system spec form and temperature model parameters to below module parameters.
- Display tracking parameters to the array metadata summary displayed when mapping fields.
- Added methods to check if the mapping process is complete. They are not currently wired up, but exist and can be used to check if each array/inverter has all required fields mapped.
- Added "time" as a required field. I'm not sure if we'd simply want to assume time is the first column but it seems like it might be a good idea to force the user to make that mapping.

Here are updated screenshots of the workflow kickoff page and the mapping page, this time with only one array.
workflow_kickoff
single_array

wholmgren:
Looking good. Minor items on the workflow kick off page:

* add "and irradiance" so block 3, item 2 reads: Calculate cell temperature from module temperature and irradiance in my data
* block 3, item 3 add comma after air temperature

We discussed breaking up the "I have data for" into a weather section (system, inverter, and array options) and a AC power section (system and inverter options). 

I'm wondering if "the entire system" will be confusing. It's the sum of the inverters (a single time series), but a user might read it as "all of the constituent parts of the system" (many time series). 

> Added "time" as a required field. I'm not sure if we'd simply want to assume time is the first column but it seems like it might be a good idea to force the user to make that mapping.

files created with pandas `to_csv` methods often don't have a name for the index column. Will that cause a problem?

lboeman:
> We discussed breaking up the "I have data for" into a weather section (system, inverter, and array options) and a AC power section (system and inverter options).
> 

Yes. The "Calculate" workflows don't require power data, so I haven't included that here. Once this is at an acceptable place, fleshing out the other workflow pages will be pretty quick.

Though this does make me think that the column mapping design needs to be modular enough to use the partial weather mapping or a power mapping for a system. I plan to get the mappings working on the front end before committing to a schema for them and implementing on the API side. Now that there's a more concrete jobs schema, I think there's a more obvious way to do it.

> I'm wondering if "the entire system" will be confusing. It's the sum of the inverters (a single time series), but a user might read it as "all of the constituent parts of the system" (many time series).
> 

Good point. Maybe this should be something like:
```
How is your weather data broken up?
- System wide measurements.
- Measurements at each inverter.
- Measurements at each array.
```
But maybe the term "measurements" isn't great for predicted data. 


> > Added "time" as a required field. I'm not sure if we'd simply want to assume time is the first column but it seems like it might be a good idea to force the user to make that mapping.
> 
> files created with pandas `to_csv` methods often don't have a name for the index column. Will that cause a problem?

Oof, yeah. I think there are two ways of handling that.
- A special "use first column" option for the time field that maps to a "" first csv header.
- Presenting columns without a header as the "Column 1", "Column 2". This would cover a case where data has multiple empty columns. Although that does open it up to the possibility for column name collision.

wholmgren:
> The "Calculate" workflows don't require power data, so I haven't included that here.

Right. Sorry.

> But maybe the term "measurements" isn't great for predicted data.

Yeah, probably better to avoid. What about

```
I will provide weather data as:

- one set for the entire system
- one set for each inverter
- one set for each array
```

@cwhanse do you have any thoughts on how to present that option more clearly?

> I think there are two ways of handling that.

No preference on solutions for the unnamed first column at this point.

cwhanse:
Maybe
```
- one set for each inverter and its associated arrays
```

The other two options are clear to me

lboeman:
I've updated the section prompting for data granularity to:
Screenshot from 2020-12-21 10-19-11


I think this is in a decent spot for mock up purposes. I think next steps are to refactor to generalize the prompts and mapping code to work for all the cases in the jobs workflow @alorenzo175 is developing in #70. Refactoring should probably be done in separate PRs. Merging this is going to drop the project test coverage considerably, but I think time shouldn't be devoted to thorough testing until refactoring.
Is there anything else that I should add here for comment before we merge this? Once it's merged it will be available at dev.solarperformanceinsight.org if there are any additional issues.

dashboard/src/views/Calculate.vue
```diff
@@ -0,0 +1,267 @@
...
+              />
+              <label for="standard">
+                global horizontal (GHI), direct normal (DNI), and diffuse
+                horizontal (DHI) irradiance.
+              </label>
+              <br />
+              <input id="poa" value="poa" type="radio" v-model="weather_type" />
+              <label for="poa">
+                global plane of array (POA global), direct plane of array(POA
+                direct), and diffuse plane of array(POA diffuse) irradiance.
```
cwhanse, line 58:
```suggestion
                global plane of array (POA global), direct plane of array (POA
                direct), and diffuse plane of array (POA diffuse) irradiance.
```

dashboard/src/views/Calculate.vue
```diff
@@ -0,0 +1,267 @@
...
+                v-model="workflow"
+              />
+              <label for="expected">
+                actual weather data during system operation.
+              </label>
+              <br />
+            </div>
+          </div>
+          <div class="my-1">
+            My data files includes:
```
cwhanse, line 41:
```suggestion
            My data file includes:
```

I'm assuming a single file is upload?

PR #70, opened 2020-12-16
Add job schemas and endpoints

alorenzo175:


api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +549,291 @@ class UserInfo(StoredObject):
...
+            self._weather_types = (JobDataTypeEnum.actual_weather,)
+            self._performance_types = (JobDataTypeEnum.actual_performance,)
+
+
+JOB_PARAMS_EXAMPLE = dict(
+    system_id=SYSTEM_ID,
+    job_type=dict(
+        compare="predicted and actual performance",
+        performance_granularity="inverter",
+    ),
```
alorenzo175, line 696: still need to handle use case 2D

alorenzo175:
@lboeman can you take a look at the models when you get a chance and let me know what you think?

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +549,291 @@ class UserInfo(StoredObject):
...
+    # in principle, these both could be on a per model chain/inverter basis,
+    # but easier to just keep everything the same for the system
+    irradiance_type: IrradianceTypeEnum
+    temperature_type: TemperatureTypeEnum
+
+    class Config:
+        schema_extra = {"example": JOB_PARAMS_EXAMPLE}
+
+
+JOB_DATE_META_EXAMPLE = dict(
```
lboeman, line 725:
```suggestion
JOB_DATA_META_EXAMPLE = dict(
```

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +549,291 @@ class UserInfo(StoredObject):
...
+    schema_path: str = Field(
+        ..., description="Relative to PV system definition, i.e. /inverters/0/arrays/0"
+    )
+    type: JobDataTypeEnum
+
+
+class JobDataMetadata(JobDataItem):
+    filename: Optional[str] = Field(..., description="Filename of file uploaded")
+    data_format: Optional[str]
+    present: bool = Field(False, description="If the data has been uploaded or not")
```
lboeman, line 744:
```suggestion
    present: bool = Field(False, description="If the data has been uploaded or not")
    
    class Config:
        schema_extra = {"example": JOB_DATA_META_EXAMPLE}
```
I think with the example above this got left out?

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +549,291 @@ class UserInfo(StoredObject):
...
+                    "status": "incomplete",
+                    "last_change": "2020-12-01T02:02:00+00:00",
+                },
+                "data_objects": [
+                    {
+                        "object_id": "6b61d9ac-2e89-11eb-be2a-4dc7a6bcd0d9",
+                        "object_type": "job_data",
+                        "created_at": "2020-12-01T01:23:00+00:00",
+                        "modified_at": "2020-12-01T01:23:00+00:00",
+                        "definition": JOB_DATE_META_EXAMPLE,
```
lboeman, line 835: Maybe I was off base with my comments above, feel free to disregard.

lboeman (commented):
I think this looks really good. It'll be easy to parse and build on the front end. I think the only thing missing is a JobStatusEnum but I know that's probably easier to define after this is more situated.

alorenzo175:
> I think this looks really good. It'll be easy to parse and build on the front end. I think the only thing missing is a JobStatusEnum but I know that's probably easier to define after this is more situated.

Thanks. Good call about the JobStatusEnum

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +552,371 @@ class UserInfo(StoredObject):
...
+        ...,
+        description="End (exclusive) of the time range that data will be uploaded for",
+    )
+    step: dt.timedelta = Field(
+        ..., description="Time step between each data point from 1 to 60 minutes"
+    )
+    timezone: Optional[str] = Field(
+        ...,
+        description="Timezone data will be converted to before computation. "
+        "Unlocalized data will be localized to this timezone. If this is null, "
```
cwhanse, line 577:
```suggestion
        "Unlocalized data will be localized to this timezone. If timezone is null, "
```

Probably should include an example string, e.g. 'Etc/GMT+5'. I usually have to look up the pytz timezone names.

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +552,371 @@ class UserInfo(StoredObject):
...
+
+    start: dt.datetime = Field(
+        ..., description="Start of the time range that data will be uploaded for"
+    )
+    end: dt.datetime = Field(
+        ...,
+        description="End (exclusive) of the time range that data will be uploaded for",
+    )
+    step: dt.timedelta = Field(
+        ..., description="Time step between each data point from 1 to 60 minutes"
```
cwhanse, line 572:
```suggestion
        ..., description="Time step between each data point in whole minutes, from 1 to 60 minutes"
```

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +552,371 @@ class UserInfo(StoredObject):
...
+
+class JobTimeindex(BaseModel):
+    """Parameters for a time index that all data uploads must conform to.
+    Data is assumed to time-averaged and closed and labeled at the left endpoint, i.e.
+    a datapoint at 23:00 of data with a 1 hour time step is assumed be the
+    average of data from 23:00 to 23:59.
+    """
+
+    start: dt.datetime = Field(
+        ..., description="Start of the time range that data will be uploaded for"
```
cwhanse, line 565: I would add an example or two of acceptable formats.

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +552,371 @@ class UserInfo(StoredObject):
...
+            raise ValueError("Could not infer timezone")
+        return values
+
+    @validator("step")
+    def check_step(cls, v):
+        secs = v.total_seconds()
+        if secs < 60:
+            raise ValueError("The minimum time step is 1 minute")
+        elif secs > 3600:
+            raise ValueError("The maximum time step is 1 hour")
```
cwhanse, line 624:
```suggestion
            raise ValueError("The maximum time step is 60 minutes")
```

api/solarperformanceinsight_api/models.py
```diff
@@ -548,3 +552,371 @@ class UserInfo(StoredObject):
...
+        return v
+
+    @validator("timezone")
+    def check_tz(cls, v):
+        if v is not None and v not in pytz.all_timezones:
+            raise ValueError("Unrecognized timezone")
```
cwhanse, line 630: Maybe

```suggestion
            raise ValueError("Unrecognized timezone, timezone must be one of pytz.all_timezones")
```

api/solarperformanceinsight_api/models.py
alorenzo175: I've updated the description to describe the formats. Most users will likely interact through the API docs which render an example like below
image

alorenzo175: and if you want to click through to view the schema
image

db/migrations/20201214175739_jobs_table.sql
```diff
@@ -0,0 +1,356 @@
...
+        select bit_and(present) from job_data where job_id = jobid
+      ) = 1;
+      if allnamed then
+        return 'prepared';
+      else
+        return 'incomplete';
+      end if;
+    end if;
+    return status;
+  end;
```
lboeman, line 74: Do we need a "checkout job" type procedure for marking the job as "running" when a worker process picks it up to compute? I'm not sure if that'd be something we compute on the fly like we're doing here, but it might be an additional condition for disallowing `add_job_data` and `queue_job`.

db/migrations/20201214175739_jobs_table.sql
```diff
@@ -0,0 +1,356 @@
...
+  definition JSON not null,
+  status enum('created', 'queued', 'complete', 'error') default 'created',
+  created_at timestamp not null default current_timestamp,
+  modified_at timestamp not null default current_timestamp on update current_timestamp,
+  primary key (id),
+  key jobs_user_id_key (user_id),
+  foreign key (user_id)
+  references users(id)
+  on delete cascade on update restrict
+) engine=innodb row_format=compressed;
```
lboeman, line 15: Are we still working under the idea that jobs would have a set lifetime? Maybe we need an expiration column here.

lboeman (commented):
I think this looks great. A couple of minor comments on the database stuff.

db/migrations/20201214175739_jobs_table.sql
alorenzo175: My thinking was that the actual status that gets returned from the API will query both the db and the queueing system. I think we probably first check the db and if `status =='queued'`, see if the job is actually running to return 'running'. Otherwise, return the db status. This way there is no state where the job started, updated the db with 'running', and then failed in some way but never updated the db. Maybe a helper script/cron is needed to check for queued jobs that have failed running for some reason but haven't updated their status to 'error'. 

I probably should have renamed `check_job_queued`, but it checks if status is anything after `created` in the status enum so `add_job_data` fails appropriately.

db/migrations/20201214175739_jobs_table.sql
alorenzo175: Yeah I'm not sure exactly how that will work, but not a bad idea. I still think I'd rather add the column later if necessary.

db/migrations/20201221161319_job_example.sql
```diff
@@ -0,0 +1,258 @@
...
+      @jobdataid1, @jobid, '/inverters/0', 'actual performance data', @extime, @extime);
+  update job_data set present = true, data = 'binary data blob', format = 'application/vnd.apache.arrow.file', filename = 'inverter_0_performance.arrow', modified_at = @modtime  where id = @jobdataid1;
+end;
+
+create procedure add_example_data ()
+  modifies sql data
+begin
+  CALL _add_example_data_0;
+  CALL _add_example_data_1;
+end;
```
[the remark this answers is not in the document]
alorenzo175: yes, should have thought of it before

db/migrations/20201214175739_jobs_table.sql
lboeman: Cool that makes sense, less state management to handle. :+1:

lboeman (approved):

Issue #63, opened 2020-12-12
options for POA, effective irradiance, module temperature, and cell temperature


wholmgren:
pvlib's `ModelChain` provides a handful of methods and data inspection for handling different input data. This issue is meant to summarize the options/behavior and determine requirements/priorities for the Solar Performance Insight development.

| Method | Required | Optional |
| -------- | --------| ----------|
| [`ModelChain.run_model`](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.modelchain.ModelChain.run_model.html) | `dni, ghi, dhi` | `temp_air, wind_speed, cell_temperature, module_temperature` |
| [`ModelChain.run_model_from_poa`](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.modelchain.ModelChain.run_model_from_poa.html#pvlib.modelchain.ModelChain.run_model_from_poa) | `poa_global, poa_direct, poa_diffuse` | `temp_air, wind_speed, cell_temperature, module_temperature` |
| [`ModelChain.run_model_from_effective_irradiance`](https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.modelchain.ModelChain.run_model_from_effective_irradiance.html#pvlib.modelchain.ModelChain.run_model_from_effective_irradiance) | `effective_irradiance` | `temp_air, wind_speed, cell_temperature, module_temperature` |

Certainly need to support `run_model`. Likely need to support `run_model_from_poa`. Is `run_model_from_effective_irradiance` of interest to this community or is it more likely to confuse users?

The new ModelChain/PVSystem/Array architecture forces us to use the same ModelChain method and required inputs for all Arrays within an Inverter. We should extend that to the Inverter level in the Insight code. I think the pvlib architecture also  requires the same optional inputs for each array, but this is not clearly documented.

[...]

PR #78, opened 2020-12-23
add expected data_columns to job metadata

alorenzo175:
This will allow us to more easily validate uploaded job data and indicates to users/frontend what data columns are required. 
Follows from discussion in #63. From my reading the conclusion is that air temp/wind speed will only be used  when not providing cell or module temperature, so I don't include them in the `data_columns`.  Is that right @wholmgren?

The JSON from a GET job now looks like 
image


api/solarperformanceinsight_api/models.py
```diff
@@ -864,12 +868,31 @@ class Job(BaseModel):
...
+        if self.parameters.irradiance_type == IrradianceTypeEnum.effective:
+            cols += ["effective_irradiance"]
+        elif self.parameters.irradiance_type == IrradianceTypeEnum.poa:
+            cols += ["poa_global", "poa_direct", "poa_diffuse"]
+        else:
+            cols += ["ghi", "dni", "dhi"]
+        if self.parameters.temperature_type == TemperatureTypeEnum.module:
+            cols += ["module_temperature"]
+        elif self.parameters.temperature_type == TemperatureTypeEnum.cell:
+            cols += ["cell_temperature"]
```
wholmgren, line 892: I'd swap the order so that it's consistent with the irradiance enum ordering of least processed to most processed.

wholmgren (commented):
> From my reading the conclusion is that air temp/wind speed will only be used when not providing cell or module temperature, 

That's correct. 

> so I don't include them in the data_columns

I don't think I understand the `data_columns` well enough to comment on this part.

lboeman (approved):

Issue #76, opened 2020-12-21
add table or time series plot to data upload page


wholmgren:
A table showing the first ~5 data rows or time series plots showing the first few days of data could provide the user with a nice visual cue that they're mapping the right columns to the right variables. 

I'd lean towards the table initially because I think it will be easier and it only requires reading a handful of lines. The table would also be helpful for indexing the row with the column names (e.g. with highlight it or put an arrow next to it).

follow up to #68.

wholmgren:
closed by #123

PR #79, opened 2021-01-06
Update workflows to utilize jobs API

lboeman:


dashboard/src/utils/transposeCSV.ts
```diff
@@ -0,0 +1,14 @@
+import csvToJson from "csvtojson";
+
+export default async function(csvString: string) {
+  /* Parses a csv file into an object mapping csv headers to column data for
+   * easier creation of partial csv files.
+   */
+  const csvJson = await csvToJson({ checkColumn: true }).fromString(csvString);
```
alorenzo175, line 7: have you looked at papaparse https://www.npmjs.com/package/papaparse? seems like it may be smaller w/ no dependencies and just as fast. Using the preview option would likely be helpful for getting the headers and the first few lines to display, and then finish parsing only when actually posting the data. The metadata/errors during parsing might be helpful, and you can probably get rid of `mapToCSV` using `unparse` and specifying the columns you want to keep.

alorenzo175 (commented):
Seems like we're getting to a point where more integration testing with the API would be useful?

dashboard/src/utils/transposeCSV.ts
lboeman: Ah, yes. I had looked at it but only when considering building the arrow tables. But now that this is using purely csv papaparse seems like a better alternative.  Thanks.

lboeman:
> Seems like we're getting to a point where more integration testing with the API would be useful?

I think we're almost there. I have two PRs before I want commit to adding those tests. 

One to update the display of the job pages and generalize the JobHandler code to build all job types.

Another to implement the other jobs.

I think that will make for a relatively stable front end to write tests against. Do you have any preferences for which testing framework to use? I've taken a look at cypress.io that has an official docker image that seems like it could be useful for adding those tests to some ci. But okay just to use selenium for python and stuff. It'd be new to me either way.

I've added Papaparse here in place of csvtojson, but retained mapToCSV to do the header conversion before post. I've left the current behavior of storing the parsed csv for now, and I'll update that handling when  #76 is implemented.

alorenzo175:
Sounds good. No preference on a testing framework; I'd agree that cypress might be best.

dashboard/package.json
```diff
@@ -9,11 +9,14 @@
...
   },
   "dependencies": {
+    "@auth0/auth0-spa-js": "^1.13.3",
     "@sentry/browser": "^5.27.6",
     "@sentry/integrations": "^5.27.6",
-    "@auth0/auth0-spa-js": "^1.13.3",
+    "@types/papaparse": "^5.2.4",
     "ajv": "^7.0.0-beta.6",
     "core-js": "^3.6.5",
+    "csvtojson": "^2.0.10",
```
alorenzo175, line 18: no longer required, right?

alorenzo175 (approved):

PR #80, opened 2021-01-06
Validate job data uploads

### LINKED ISSUE #64: interval label, type requirements
We need to decide if/how we will account for interval label and interval value type. From [solar forecast arbiter definitions](https://solarforecastarbiter.org/definitions/#forecastattrs)

* Interval length - The length of time that each data point represents, e.g. 5 minutes, 1 hour.
* Interval label - Indicates if a time labels the beginning or the ending of an interval average, or indicates an instantaneous value, e.g. beginning, ending, instant
* Interval value type - The type of the data in each interval of the forecast, e.g. mean, maximum, median.

Do we have restrictions on interval length? Longest interval of 1 hour? Is there a need to support or restrict intervals shorter than 1 minute? We'll have an upper limit on upload data size, so shorter intervals may run into restrictions on total length.

Can we require a particular interval label? I doubt it. Can we treat them all as instantaneous? Probably fine for 5 - 15 minute intervals, but we might need to do better for 1 hour intervals. What's the scope for 1.0 vs someday? If we perform upsampling and averaging, what's the scope for Insight vs. pvlib? Relevant discussion in the [pvlib google group](https://groups.google.com/g/pvlib-python/c/b4uMI3v3qOc).

Interval value type. Only mean or instantaneous. I don't see an application for anything else.

Also related: we should require that all time series data has the same properties.

### LINKED ISSUE #77: Validated posted job data
validate:
- correct columns
- time index matches expected?

alorenzo175:
closes #64 closes #77

Validates uploaded job data to: 
- conform to the times as specified for the job including adding nans if appropriate
- fail if more than 10% of times are missing
- fail if more than 10% of times are extra (compared to time index spec for job)
- fail if any columns are missing
- only save required columns
- save as arrow file format to database
- accepts either arrow file format data or csv
- validates times and numeric columns

Returns some information about the values that were uploaded like how many were nan or how many and which times were missing.

Also allows the data to be retrieved in either arrow file format or as a csv.


alorenzo175:
@wholmgren any other validation suggestions or is anything too strict?

wholmgren:
Generally looks good. I don't see anything missing right now but maybe something will come up when we can start using the upload pages.

> fail if more than 10% of times are missing

Seems reasonable to me but worth checking with @cwhanse. I can imagine making this user adjustable but we're in a time crunch and I think that should wait.

> fail if more than 10% of times are extra (compared to time index spec for job)

Does this mean outside the slice [start:end]? Assuming that we only use the start to end, I'm not sure why we'd fail in this case.

api/solarperformanceinsight_api/utils.py
```diff
@@ -0,0 +1,197 @@
...
+        )
+    if "time" in expected:
+        if not is_datetime64_any_dtype(df["time"]):
+            raise HTTPException(
+                status_code=400,
+                detail='"time" column could not be parsed as a timestamp',
+            )
+        if not (df["time"].dt.microsecond == 0).all():  # type: ignore
+            raise HTTPException(
+                status_code=400, detail='"time" column must only have second precision'
```
cwhanse, line 113: I could see this being a warning that time is being rounded for the user, instead of an Exception.

api/solarperformanceinsight_api/utils.py
```diff
@@ -0,0 +1,197 @@
...
+            )
+        # check for duplicates
+        extra_times = len(df["time"]) - len(df["time"].unique())
+        if extra_times != 0:
+            raise HTTPException(
+                status_code=400,
+                detail=f'"time" column has {extra_times} duplicate entries',
```
cwhanse, line 120: Agree with this one.

cwhanse:
> > fail if more than 10% of times are missing
> 
> Seems reasonable to me but worth checking with @cwhanse. I can imagine making this user adjustable but we're in a time crunch and I think that should wait.

Agree. It's a reasonable threshold, not sure I'd make it adjustable. The better solution is for the user to resample/interpolate the data and resubmit.

> > fail if more than 10% of times are extra (compared to time index spec for job)
> 
> Does this mean outside the slice [start:end]? Assuming that we only use the start to end, I'm not sure why we'd fail in this case.

If extra times are those outside of [start:end], I'd just warn and move on with the analysis.

lboeman (approved):

alorenzo175:
> > > fail if more than 10% of times are extra (compared to time index spec for job)
> > 
> > 
> > Does this mean outside the slice [start:end]? Assuming that we only use the start to end, I'm not sure why we'd fail in this case.
> 
> If extra times are those outside of [start:end], I'd just warn and move on with the analysis.

Yes, extra times are those outside of [start:end). My thinking here was that I wanted to clearly alert the user that the frequency of the results they will get is not what they uploaded, i.e. a user specifies a job with step of 1h but uploads 1 minute data, expects 1 minute performance output, but gets 1 h instead. Since this is all through interaction through the API, I don't know of a great way to warn the user besides the response body which is easily ignored if the status code is 200. An alternative approach would be to add a query parameter like `ignore_extra_times=true` that would accept any extra times and carry on, but return a 400 to unaware users that don't include it.

Thoughts on query parameter vs just moving on w/ counts/extra times in the response body?

api/solarperformanceinsight_api/utils.py
alorenzo175: Yeah, this I doubt many users would upload microseconds in the first place. I removed this check and just round them off when reindexing

wholmgren:
>i.e. a user specifies a job with step of 1h but uploads 1 minute data, expects 1 minute performance output, but gets 1 h instead. 

Maybe that needs a more explicit test and warning.

```
In [6]: index = pd.DatetimeIndex(['20200101 0000', '20200101 0100', '20200101 0200', '20200101 0400'])

In [7]: index.to_series().diff().mode()
Out[7]:
0   0 days 01:00:00
dtype: timedelta64[ns]

In [9]: offset = pd.tseries.frequencies.to_offset(index.to_series().diff().mode().iloc[0])

In [10]: offset
Out[10]: <Hour>

In [11]: expected = pd.date_range(start='20200101 0000', end='20200101 0400', freq='1h')

In [12]: offset == expected.freq
Out[12]: True
```

In any case I don't think a query parameter is needed at this time.

alorenzo175:
Ok, no longer error when there are extra rows, and now include the expected period and period of uploaded data in the result
image

PR #81, opened 2021-01-08
add timezone parameter endpoint to get list of valid timezones

alorenzo175:


api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -50,3 +51,9 @@ def get_sandia_inverter(
     return models.SandiaInverterParameters(
         **sandia_inverter_params[inverter_name].to_dict()
     )
+
+
+@router.get("/timezones", response_model=List[str])
+def get_timezones() -> List[str]:
+    """Get all recognized timezones"""
+    return pytz.all_timezones
```
wholmgren, line 59: this is fine but consider `common_timezones` instead.

also worth noting that we used US + GMT offsets in the solar forecast arbiter

https://github.com/SolarArbiter/solarforecastarbiter-dashboard/blob/cd0538bb82f071c5550ed5fa8825bf9c6e6f1e7c/sfa_dash/template_globals.py#L19-L20

alorenzo175: `common_timezones` does not include the GMT offsets which are certainly useful

lboeman (approved):
Endpoint looks good.

PR #82, opened 2021-01-09
Job Setup layout update and refactor

lboeman:
Updates the layout of job setup pages and refactors job related vue components.
Screenshot from 2021-01-08 17-39-33

- Changes the `JobHandler` to also handle job kick off.
- Moved the job setup/definition code into a parameters directory. With a `JobParams` component responsible for the common job parameters and a component for the each specific set of `job_type` parameters that can be nested in `JobParams` for a complete job object..
-Generalizes the Weather upload and weather mapper component to any csv upload (this is largely a naming update).
- Creates one "Step" in the side bar for each type of data found in the job's `data_objects` (the set of the data_objects' types).


alorenzo175 (approved):
This is great for now. In the future we might fix the routing between the left tabs and maybe a warning when leaving the job before computing and deleting the job if possible when one does leave the page w/o computing

PR #83, opened 2021-01-11
fix csv types and add build step as part of action testing

alorenzo175:


dashboard/src/components/jobs/WeatherUpload.vue
```diff
@@ -108,7 +108,8 @@ export default class WeatherUpload extends Vue {
       console.log("Bad csv");
     } else {
       this.csvData = parsingResult.data;
-      this.headers = parsingResult.meta.fields;
+      const headers = parsingResult.meta.fields;
+      this.headers =  headers ? headers: [];
```
alorenzo175, line 112: not sure this is the proper way to set `this.headers = []` if headers is undefined otherwise set `this.headers = headers`

lboeman: Papa parse should generate an error when there are no headers to parse here and hit the block above. Is this something you've triggered or a type warning?

lboeman (approved):
Thanks for adding this.

dashboard/src/components/jobs/WeatherUpload.vue
alorenzo175: type error when building

PR #85, opened 2021-01-14
Add Cypress integration testing

lboeman:
Adds a workflor for https://github.com/cypress-io/github-action to run [cypress](https://www.cypress.io/) end to end tests against the dashboard backed by the api. Copies the API test steps for setup, so the migration/api install step in this workflow will need to stay synced with those. As of opening, just hits the home page unauthenticated and then finishes. Working on adding testing for authenticated pages.


lboeman:
@alorenzo175 This should serve as the start of real integration tests. I got this to work in my fork, but here it seems unable to access secrets. Any ideas?

lboeman:
> @alorenzo175 This should serve as the start of real integration tests. I got this to work in my fork, but here it seems unable to access secrets. Any ideas?

Ah, scratch that. Secrets aren't accessible from PR actions.

alorenzo175:
> > @alorenzo175 This should serve as the start of real integration tests. I got this to work in my fork, but here it seems unable to access secrets. Any ideas?
> 
> Ah, scratch that. Secrets aren't accessible from PR actions.

Yeah for better or worse, I think all of those things are already in the api code (conftest.py and __init__.py) so you can just hardcode them in the workflow or elsewhere. I don't think you should need the client secret at all.

alorenzo175 (approved):
Is it worth saving screenshots as artifacts? https://github.com/cypress-io/github-action#artifacts

lboeman:
> Is it worth saving screenshots as artifacts? https://github.com/cypress-io/github-action#artifacts

Yeah, good call.

PR #87, opened 2021-01-14
Add functionality for job results

alorenzo175:
only have list/get results endpoints since it's the expectation that a job worker uses the storage interface directly to add results for a job


lboeman (approved):
This looks good to me. I didn't expect the results to be accessible like this, but I like it, and it gives us a little leeway on implementation without changing the API. :+1:

PR #89, opened 2021-01-16
Add RQ integration

alorenzo175:


lboeman (approved):
Seems like a pretty reasonable interface to me. Just to make sure I follow. Is there some worker process that would run and call the housekeeping functions in `QueueManager`?

alorenzo175:
> Is there some worker process that would run and call the housekeeping functions in `QueueManager`?

That's right; plan to add the loop in a separate PR

Issue #52, opened 2020-12-09
Only allow one array per inverter for new system


alorenzo175:
and probably auto-populate the form with one inverter and possibly one array. Eventually we'll support multiple arrays per inverter, so maybe just gray out the Add Array button or alert when adding more than one?

wholmgren:
The pvlib functionality for multiple arrays is almost ready to merge into master. One thing that came up in review is that single axis trackers will be restricted to a single array for now. I'm -1 on hacking around that limitation in SPI. It would be far better to improve the situation in https://github.com/pvlib/pvlib-python/issues/1109. In the meantime, we might need to consider graying out that button if a tracker is provided.

wholmgren:
set requirements file to use pvlib master for now and remove this limitation in the api. we can also cut pvlib 0.9 pre-releases if that's preferable.

lboeman:
After #92 We are at a point where multiple arrays are supported when all arrays have fixed tracking. The dashboard should:

- Only allow selecting between fixed/single axis tracking when one array exists on the inverter.
- Only allow users to add additional arrays (through "Add Array" or "Duplicate Array") if single axis is not selected.

alorenzo175:
dashboard limits in #94

PR #92, opened 2021-01-21
use pvlib.Arrays and lift the requirement of only a single array

alorenzo175:
partially addresses #52, but still need to avoid multiple arrays for
single axis tracking on the dashboard


api/solarperformanceinsight_api/models.py
```diff
@@ -316,7 +316,7 @@ class PVWattsInverterParameters(BaseModel):
     eta_inv_ref: float = Field(
         0.9637, description="Reference inverter efficiency, unitless"
     )
-    _modelchain_ac_model: str = PrivateAttr("pvwatts")
+    _modelchain_ac_model: str = PrivateAttr("pvwatts_multi")
```
wholmgren, line 319: I think we need to use `pvwatts` if only 1 inverter is present. `pvlib.inverter.pvwatts_multi` assumes tuple/list input and will return the wrong result if provided with only a Series. Is that doable here or not trivial? I'm guessing the latter.

I'm open to changing things in pvlib, but it's also not trivial there.

wholmgren: same for sandia_multi

alorenzo175: Do you mean calls to `ModelChain.run_model` always need input as a list (`ModelChain.run_model([array_0_data, array_1_data])`) for this to work? That is the current plan for calling run_model, even with a single array.

cwhanse: > I think we need to use pvwatts if only 1 inverter is present. 

Agree. 

The pvlib implementation could be made flexible to cover single and multiple inverters, if that becomes necessary.

wholmgren: From just looking at the code, I think `pvwatts_multi` might actually work with a single array within the `ModelChain` workflow. `test_modelchain.py::test_ac_models` asserts that `mc.results.ac` is a Series. This would not be the case if `pvwatts_multi` performed a sum over the single series.

lboeman (approved):
Looks good. Looks like for now the frontend should:
- Only allow selecting between fixed/single axis tracking when one array exists on the inverter.
- Only allow users to add additional arrays (through "Add Array" or "Duplcate Array") if single axis is not selected.

PR #93, opened 2021-01-22
Add job function for calculating performance

### LINKED ISSUE #84: Add calculate performance functions
These should take a StoredJob object and any associated job data, and calculate performance via pvlib. I think we can worry about exactly what attributes are returned and what goes into a report later.

alorenzo175:
closes #84


api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+            yield [df.copy()] * num_arrays
+    elif (
+        job.definition.parameters.weather_granularity
+        == models.WeatherGranularityEnum.inverter
+    ):
+        for i in range(num_inverters):
+            num_arrays = len(job.definition.system_definition.inverters[i].arrays)
+            data_id = data_id_by_schema_path[f"/inverters/{i}"]
+            df = _get_data(job_id, data_id, si, shift)
+            yield [df] * num_arrays
```
alorenzo175, line 105: maybe these need to be copies of df depending on if pvlib does any modifying of the frames

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
 
-def dummy_func(job, storage):
+def dummy_func(job, storage):  # pragma: no cover
     raise NotImplementedError("Job computation not implemented")
+
+
+def _get_data(
+    job_id: UUID, data_id: UUID, si: storage.StorageInterface, shift: dt.timedelta
+) -> pd.DataFrame:
+    """Get the data from the database and shift forward by `shift`"""
```
wholmgren, line 63: I'm not thrilled about the "*and* shift forward" part of this function. It means this function does more than one useful thing. It also makes the requirements for `process_single_modelchain` harder to understand - the data has to be shifted but you still have to supply `shift` so you can unshift the results? Why not do all the shifting within `process_single_modelchain`?

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+    summary_frame : pd.DataFrame
+        A frame with the AC performance result, zenith angle, and average of
+        poa_global, effective_irradiance, and cell_temperature over all arrays
+    """
+    # run chain
+    mc = getattr(chain, run_model_method)(weather_data)
+    results = mc.results
+    adjust = partial(_adjust_frame, tshift=tshift)
+    performance = adjust(results.ac, "performance")
+    summary_frame = pd.DataFrame(
```
wholmgren, line 218: I finally got that we're just computing average quantities, but between the names, the nested functions, and the out.append lines it took me awhile to see it. Some combination of the following comments would make it more clear to me...

wholmgren: call this `summary_frame` something like `weather_sum`

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+        {
+            "poa_global": 0,  # type: ignore
+            "effective_irradiance": 0,  # type: ignore
+            "cell_temperature": 0,  # type: ignore
+        },
+        index=performance.index,
+    )
+    summary_frame.index.name = "time"  # type: ignore
```
wholmgren, line 226: can this be moved to the end of the function?

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+    # make nice result output
+    num_arrays = len(mc.system.arrays)
+    out = []
+    for i in range(num_arrays):
+        weather: pd.DataFrame = adjust(  # type: ignore
```
wholmgren, line 231: call this `array_weather`

wholmgren: can we have a single adjust for the whole dataframe (before the `+=`) instead of calling it on every series?

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+                data=weather,
+            )
+        )
+    out.append(
+        DBResult(
+            schema_path=f"/inverters/{inverter_num}",
+            type="performance data",
+            data=performance,
+        )
+    )
```
wholmgren, line 254: move this below the `/=` or even lower.

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+    # mean
+    summary_frame /= num_arrays  # type: ignore
+    summary_frame.insert(0, "performance", performance)  # type: ignore
```
wholmgren, line 257: `summary_frame = pd.concat([performance, weather_avg, zenith], axis=1)` or similar

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+    tshift : dt.timedelta
+        Weather data should already be shifted forward by this amount, so final results
+        will be shifted by -1 * tshift.
+    inverter_num : int
+        Which inverter in the full system this chain is for
+
+    Returns
+    -------
+    inverter_results : List[DBResult]
+        List of DBResult objects that can be inserted into the database
```
wholmgren, line 208: probably worth saying that these results contain AC performance for the entire system and poa_global, effective_irradiance, and cell_temperature for each array.

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+            "zenith": 0,  # type: ignore
+        },
+        index=job_time_range,
+    )
+    summary.index.name = "time"  # type: ignore
+    run_model_method = job.definition._model_chain_method
+    tshift = job.definition.parameters.time_parameters.step / 2
+    chains = construct_modelchains(job.definition.system_definition)
+
+    weather_count = 0.0
```
wholmgren, line 287: why is this a float?

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+            type="performance data",
+            data=performance,
+        )
+    )
+    # mean
+    summary_frame /= num_arrays  # type: ignore
+    summary_frame.insert(0, "performance", performance)  # type: ignore
+    summary_frame.insert(
+        len(summary_frame.columns),
+        "zenith",
```
wholmgren, line 260: Should we use apparent zenith? I think that would be more correct for this application, but not much difference at 87 degrees so I'm fine if we want to keep zenith for simplicity.

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,326 @@ def run_job(job_id: UUID, user: str):
...
+        weather_count += 1
+    # keep performance as sum, but make everything else average over inverters
+    total_performance = summary.pop("performance")  # type: ignore
+    # summary up to now is sum of array-averaged weather for each inverter
+    summary /= weather_count  # type: ignore
+    # summary is now average over inverters
+    # average zenith
+    daytime = summary.pop("zenith") < 87.0  # type: ignore
+    daytime.name = "daytime_flag"  # type: ignore
+    # index of data actualy uploaded
```
wholmgren, line 310:
```suggestion
    # index of data actually uploaded
```

api/solarperformanceinsight_api/compute.py
alorenzo175: old habit to make sure result after division is float

api/solarperformanceinsight_api/compute.py
alorenzo175: This was to simplify `generate_weather_data` and later use of the list unmodified. I can see how just iterating weather data and shifting there is easier to follow

api/solarperformanceinsight_api/compute.py
alorenzo175: block above reuses the same dataframe on each iteration vs getting data for each iteration here. We won't have to worry about copying here for our current use since the later `.shift` will produce a copy

alorenzo175:
I can revert to output kW/kWh if a definite decision is made

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,338 @@ def run_job(job_id: UUID, user: str):
...
 
-def dummy_func(job, storage):
+def dummy_func(job, storage):  # pragma: no cover
     raise NotImplementedError("Job computation not implemented")
+
+
+def _get_data(
+    job_id: UUID, data_id: UUID, si: storage.StorageInterface
+) -> pd.DataFrame:
+    """Get the data from the database and shift forward by `shift`"""
```
wholmgren, line 63:
```suggestion
    """Get the data from the database."""
```

api/solarperformanceinsight_api/compute.py
```diff
@@ -30,22 +35,338 @@ def run_job(job_id: UUID, user: str):
...
+    chain : pvlib.modelchain.ModelChain
+        The chain to run
+    weather_data : List[pandas.DataFrame]
+        The data that will be passed to chain.run_model
+    run_model_method : str
+        The method of chain to used calculate results, i.e. run_model,
+        run_model_from_poa, run_model_from_effective_irradiance
+    tshift : dt.timedelta
+        Weather data should already be shifted forward by this amount, so final results
+        will be shifted by -1 * tshift.
```
wholmgren, line 199:
```suggestion
        Amount by which to shift weather data before running the chain.
        Typically half the interval length.
```

wholmgren (approved):
looks good except for minor doc changes...

Issue #88, opened 2021-01-15
dashboard review issues


wholmgren:
Here's a long list of things I encountered while playing with the dev site this morning. Many of these deserve their own issues, and maybe some of them are duplicates

- [x] tabbing should skip help boxes
      Should add `tabindex=-1` to buttons and add `aria-labelledby` attribute to input elements so that help text is accessible.
- [x] sure would be nice if we used an elevation look up service
      Can probably use a  free elevation lookup service like https://api.opentopodata.org/ for and adopt a paid service for a full fledged product. Should probably be manually activated from a button to avoid hitting free service limits.
- [x] remove the default system name. remove the default inverter/array name or change it to an auto-incremented value like "Inverter A" or "Inverter 1"
- [x] pvwatts loss parameter `Lid` should be `LID`. LID = Light Induced Degradation
- [x] clarify units throughout dashboard. (done in metadata and #148 will likely do the rest)
- [ ] maybe put another save system button at the bottom of the page
- [x] Display JSON Summary button should be a block with a collapsible arrow. Maybe "Display/Hide JSON Representation"
- [ ] can we round numbers in the inverter database? for example, `C3: -0.014925999999999998` is annoying to read. @lboeman says PR incoming soon. #164
- [ ] "duplicate" or "save as" functionality for systems (#73)
- [ ] improve api error handling (no longer remember what this specifically means)
- [x] are the pvwatts modules per string and strings options wired up in pvlib? no, this is https://github.com/pvlib/pvlib-python/issues/476
- [x] move tracking box to be above modules per string
- [x] "my weather data file includes" starts with GHI, DNI, DHI and ends with "effective irradiance", while cell temperature starts with "cell temperature" and ends with "...from irradiance, air temperature...". It would be more consistent if the cell temperature options were reversed.
- [ ] guess at column names. e.g. if a column as "time" then use it for the timestamp column. if a column contains "global" or "GHI" then use it for the GHI column. (out of scope for 1.0, no follow up issue to keep things simple)

lboeman:
>     * [ ]  clarify units throughout dashboard

For the System specification page, units are mostly found in the help text which are pulled from the API specification. Should the units be displayed next to the inputs as they are in the arbiter? 

>     * [ ]  are the pvwatts modules per string and strings options wired up in pvlib?

Looks like those options were added with the new `Array` class  in https://github.com/pvlib/pvlib-python/pull/1076 ? 

[...]

PR #94, opened 2021-01-22
System Specification Updates

lboeman:
Handles Issues brought up in #88 particular to the System Specification page.

- Tabbing should skip help boxes. Removed the help buttons from the tab index, so that they aren't accessible via keyboard. Added some ARIA attributes to describe the inputs using the help text when using screenreaders, since the buttons would no longer be accessible.
- Elevation lookup button via google maps javascript api/elevation service. Requires that an api key be specified using the VUE_APP_GOOGLE_MAPS_API_KEY environment variable. Button can be seen below (with the fun -3492 result at lat/lon 0,0). Maybe best below/next to the elevation input?  
  Screenshot from 2021-01-22 12-49-13

- Removed the default system name. Inverters and Array should be added with incrementing names e.g. "Inverter 1", "Invereter 2". 
  ~~Need to fix bug where removing and adding an inverter/array results in same name. e.g. add "Inverter 1", "Invereter 2", delete inverter 1 and add a second invereter 2.~~ This is actually less confusing then adding adding a second inverter and getting "Inverter 7" or something. 
- Moved JSON summary toggle to a collapsible block.
  Screenshot from 2021-01-22 12-47-58

- PVWatts Lid => LID. This is the only change in `models.py`

Needs tests and I'll add the follow functionality from #92 here as well.


alorenzo175 (commented):
I would prefer the elevation button next to the elevation field. 

When clicking on a single axis tracker I get this console error
```
[Vue warn]: Property or method "fieldID" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <ModelField> at src/components/ModelField.vue
       <TrackingParametersView> at src/components/model/TrackingParameters.vue
         <ArrayView> at src/components/model/Array.vue
           <ArraysView> at src/components/model/Arrays.vue
             <InverterView> at src/components/model/Inverter.vue
               <InvertersView> at src/components/model/Inverters.vue
                 <SystemView> at src/components/model/System.vue
                   <SystemSpec> at src/views/SystemSpec.vue
                     <Home> at src/views/Home.vue
                       <App> at src/App.vue
                         <Root> vue.runtime.esm.js:619
    VueJS 3
    render ModelField.vue:98
    VueJS 29
    change Array.vue:82
    VueJS 33
```

alorenzo175 (approved):

lboeman:
Updated to use https://nationalmap.gov/epqs/ over Google maps to avoid any handling of keys. Only really covers Continental U.S. but would be easy in the future to integrate a different service.

lboeman:
@alorenzo175 updated `dashboard/src/utils/elevation.ts` since approval. Still good to merge with you?

alorenzo175 (approved):

wholmgren:
@ncroft-b4 thanks for the https://nationalmap.gov/epqs/  suggestion!

Issue #97, opened 2021-01-27
UC 2C: compare expected and actual performance


alorenzo175:
#65 has gotten really crowded, so I wanted to make a dedicated issue/discussion for the backend calculation and dashboard workflow of 2C. 

I think the current dashboard form works just fine for this use case:
![image](https://user-images.githubusercontent.com/4625457/106020471-0c662300-6081-11eb-86f2-a604053c4502.png)

The user can specify any of the options for the weather data, then upload actual weather and actual performance data. The weather data and performance data must have the same time index that matches what's specified at the bottom of the form. 

The backend will first calculate the expected performance from the actual weather data. We can provide the same outputs as 1B (expected system performace, expected performance per inverter, weather (poa, effective irr., cell temp) for each array, monthly summary of energy/insolation) or limit to a subset. 

Then the backend will calculate the monthly energy from the uploaded actual performance, summing the performance across inverters if the performance data was supplied on a per inverter basis. The additional output will be a table comparing the expected monthly energy to the actual monthly energy

Month (or other bin) | Actual (kWh) | Modeled (kWh) | Difference (kWh) | Percent Actual to Modeled
-- | -- | -- | -- | --
Jan | ... | ... | ... | ...
... | ... | ... | ... | ...
Dec | ... | ... | ... | ...
Total | ... | ... | ... | ...

```python
def compare_expected_and_actual(expected_weather, actual_performance):
    assert expected_weather.index == actual_performance.index
    expected_system_performance, other_1b_results = calculate_expected_performance(expected_weather)
    expected_energy = expected_system_performance.resample('1h').mean().groupby(month).sum()
    if actual_performance_is_per_inverter:
        actual_performance = sum(actual_performance)
    actual_energy = actual_performance.resample('1h').mean().groupby(month).sum()
    return other_1b_results, comparison_table(expected_energy, actual_energy)
```

alorenzo175:
closed by #99

PR #99, opened 2021-01-28
add comparison of calculated expected performance and actual

alorenzo175:
API support for UC 2c #97


api/solarperformanceinsight_api/compute.py
```diff
@@ -369,4 +377,52 @@ def run_performance_job(job: models.StoredJob, si: storage.StorageInterface):
...
+    )
+    diff = actual_monthly_energy - expected
+    ratio = actual_monthly_energy / expected
+    comparison_summary = pd.DataFrame(
+        {
+            "actual_energy": actual_monthly_energy,
+            "expected_energy": expected,
+            "difference": diff,
+            "ratio": ratio,
+        }
```
alorenzo175, line 420: @ncroft-b4 Is this what you had in mind for the output for use case 2C (in addition to the expected performance timeseries and weather timeseries)?

ncroft-b4: Yes, that covers it. The monthly sum of insolation is also of interest, but it is not as important and its probably good to avoid displaying too much information on the output, and that can be pulled from the timeseries. So I say go with this

lboeman (approved):

api/solarperformanceinsight_api/compute.py
```diff
@@ -386,4 +394,52 @@ def run_performance_job(job: models.StoredJob, si: storage.StorageInterface):
...
+        do.object_id
+        for do in job.data_objects
+        if do.definition.type == models.JobDataTypeEnum.actual_performance
+    ]
+    months = (
+        (job.definition.parameters.time_parameters._time_range)
+        .month.unique()  # type: ignore
+        .sort_values()
+    )
+    expected = monthly_energy["performance"]  # type: ignore
```
lboeman, line 419: Actually, running this I'm hitting a KeyError here, where monthly energy is a Series named "performance" instead of a DataFrame with a "performance" column.

PR #102, opened 2021-01-28
Add job result rendering

lboeman:


lboeman:
This adds basic functionality for rendering a summary table and time series plots. This is a minimal implementation with very little styling that only includes the ability to view a single timeseries at a time.

The installation process now has a post-install step that can be run with `npm run postinstall` I've added it to the ci/dockerfile. It's needed to rename the type definitions from definitely typed (@types/plotly.js) to match plotly.js-basic-dist. Using the basic-dist provides the traces we need for simple timeseries and results in a 1.7MB built vendor.js v.s. 4MB and only adds a couple hundred lines to package-lock.json vs 3000 for the full package. 

I still need to add download functionality for results but @alorenzo175 if you could take a look at this I'd appreciate any feedback.

alorenzo175 (commented):
Good start. A few things:

- [ ] hard for me to tell exactly, but a console error of a null job result prevented the result from showing until I refreshed. Along with that is there a mechanism to check the status and then load once finished? later a updated running status would be cool
- [ ] I think we probably only want to show time series to people that want it by clicking on a button or dropdown or something
- [ ] in addition to the results, we should provide links to get back the data you uploaded as processed by the system (/jobs/id/data/dataid). maybe this would be on the submit calc page?

lboeman:
>     * [ ]  in addition to the results, we should provide links to get back the data you uploaded as processed by the system (/jobs/id/data/dataid). maybe this would be on the submit calc page?

I agree. I think the calculation page  is a good place for that, I had thought about adding options to allow users to plot/preview the uploaded data before submitting. Do you think that'd also be useful or should it just be a table of uploaded data with download links?

Thanks for the review, I'll work out that bug, get the running status working and make timeseries plots a user-activated thing.

alorenzo175:
> I agree. I think the calculation page is a good place for that, I had thought about adding options to allow users to plot/preview the uploaded data before submitting. Do you think that'd also be useful or should it just be a table of uploaded data with download links?

Right, I forgot about the preview before submit. From what I heard, people like the table option to download, and maybe we can add a display as plot button later

wholmgren:
>  I think we probably only want to show time series to people that want it by clicking on a button or dropdown or something

As discussed previously, we definitely don't want to download all time series data on the initial page load, but I think we should start with a time series plot with total AC power and let users modify from there. Maybe the group will say that's not important, but we should be prepared for either request after they can play with it.

lboeman:
>     hard for me to tell exactly, but a console error of a null job result prevented the result from showing until I refreshed. Along with that is there a mechanism to check the status and then load once finished? later a updated running status would be cool

Fixed the null job error. Instead of passing just the job's id. The whole object is passed to the `JobResult` component. When the job is not "complete" and the component is active, it emits an event to update the job every second until the jobStatus is "completed", showing a different message for "queued" and "running". 


>      I think we probably only want to show time series to people that want it by clicking on a button or dropdown or something

Turned the options into a select list that defaults to "Select a timeseries result" and only displays a plot when a selection is made.

>     in addition to the results, we should provide links to get back the data you uploaded as processed by the system (/jobs/id/data/dataid). maybe this would be on the submit calc page?

Just created an issue for this. Created a `downloadFile` utility function/module that takes a blob and a name to make implementation pretty simple.

Also added the ability to download the results.

lboeman:
I think this is in a good spot for demo purposes and getting feedback before committing to a design and adding tests.

alorenzo175 (approved):
Looks good for demoing to me. Maybe make an issue to come back and write tests

PR #118, opened 2021-02-03
Fix job result timeseries selection.

### LINKED ISSUE #116: More careful job timeseries selection
If a user names two arrays the same thing, they show up under the same name in the timeseries picker and seems like the data doesn't change. Might need to have some extra information like "Array 1 weather data (/inverters/0/array/1)" and "Array 1 weather data (/inverters/1/arrays/0)" possibly indexed in js by the path + type with a name mapping for selection

lboeman:
closes #116 
Appended the schema path to the end of each timeseries result to differentiate system components with the same name.

The timeseries plotting component was watching for updates to the Arrow table and setting the currently selected column to the first variable in the table, changing columns triggered the redraw through a watch function on column. When switching between similar results, the column was never actually updated.
I've updated the function that watches for arrow table changes to trigger a redraw if the column has not changed.

### DIFF
--- a/dashboard/src/components/jobs/JobResults.vue
+++ b/dashboard/src/components/jobs/JobResults.vue
@@ -149,7 +149,7 @@ export default class JobResults extends Vue {
           this.system,
           result.definition.schema_path
         );
-        const label = `${systemComponent.name} ${result.definition.type}`;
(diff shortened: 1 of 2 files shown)


alorenzo175 (approved):

alorenzo175:
good to merge?

PR #123, opened 2021-02-04
Update csv upload

### LINKED ISSUE #114: Improve CSV uploads
- [ ] better error validation from the API
- [ ] configurable number of lines to skip
- [ ] spinner when first processing csv
- [ ] spinner when uploading to api

lboeman:
closes #114 and adds a CSV preview table at the top of the page.


alorenzo175:
very cool!

lboeman:
Couple extra things to fix before this is ready: 

- [ ] Make the header/data rows adjustable, so if you get them wrong you don't have to completely restart the process.
- [ ] Add "Not Mapped" placeholder to the mapped variable row.
- [ ] Ensure errors are cleared if  different file is selected.

alorenzo175 (commented):
There's a problem with the column mapping. For example, upload a file with some extra columns, then make all selections. Then try and change time to another column. For me, the "time" column originally selected remains disabled instead of being released back. It also allowed me to map the same column to multiple variables

alorenzo175 (approved):

PR #124, opened 2021-02-05
update for pvlib 0.9.0a4

### LINKED ISSUE #109: update _modelchain_ac_model for multi array inverters
https://github.com/pvlib/pvlib-python/pull/1147 and https://github.com/pvlib/pvlib-python/pull/1150 changed the object oriented api for the new multi-array inverter functions. Now you just pass "sandia" or "pvwatts" and the objects determine which function they should call. So SPI needs to update `PVWattsInverterParameters._modelchain_ac_model` and `SandiaInverterParameters._modelchain_ac_model`. The change is available in `v0.9.0-alpha.2`

### LINKED ISSUE #119: Single axis tracker systems with ghi, etc might fail to calculate
Probably related to the incorrect handling of singleton arrays overall.

Sentry Issue: [RQ-WORKER-3](https://sentry.io/organizations/solar-performance-insight/issues/2196812857/?referrer=github_integration)

```
ValueError: Length of passed values is 1, index implies 16800.
(8 additional frame(s) were not displayed)
...
  File "pvlib/irradiance.py", line 781, in haydavies
    AI = dni / dni_extra
  File "pandas/core/ops/common.py", line 65, in new_method
    return method(self, other)
  File "pandas/core/ops/__init__.py", line 345, in wrapper
    return left._construct_result(result, name=res_name)
  File "pandas/core/series.py", line 2769, in _construct_result
    out = self._constructor(result, index=self.index)
  File "pandas/core/series.py", line 313, in __init__
    raise ValueError(

Error for job UUID('d31a5277-663d-11eb-960d-0afeb1772d65')
```

alorenzo175:
a4 because https://github.com/pvlib/pvlib-python/pull/1162 is required
along with a tbd issue for computing from effective irradiance

closes #119 closes #109

also adds real tests of the pvlib modelchain to catch errors there

### DIFF
--- a/api/solarperformanceinsight_api/compute.py
+++ b/api/solarperformanceinsight_api/compute.py
@@ -259,11 +259,13 @@ def process_single_modelchain(
     # mean
     weather_avg: pd.DataFrame = weather_sum / num_arrays  # type: ignore
     adjusted_zenith: pd.DataFrame
-    # not calculate if effective irradiance is provided
+    # not calculated if effective irradiance is provided
     if results.solar_position is not None:
         adjusted_zenith = adjust(results.solar_position[["zenith"]])  # type: ignore
     else:
         # calculate solar position making sure to shift times and shift back
+        # modelchain passes through air temperature and pressure, but that only
+        # affects apparent_zenith
         adjusted_zenith = adjust(
             mc.location.get_solarposition(
                 weather_avg.index.shift(freq=tshift)  # type: ignore
--- a/api/solarperformanceinsight_api/compute.py
+++ b/api/solarperformanceinsight_api/compute.py
@@ -242,7 +242,7 @@ def process_single_modelchain(
         array_weather.loc[:, "poa_global"] = _get_index(
             results, "total_irrad", i
         ).get(  # type: ignore
-            "poa_global"
+            "poa_global", float("NaN")
         )
         array_weather.loc[:, "cell_temperature"] = _get_index(
             results, "cell_temperature", i
--- a/api/solarperformanceinsight_api/compute.py
+++ b/api/solarperformanceinsight_api/compute.py
@@ -222,8 +222,8 @@ def process_single_modelchain(
     ).to_frame()  # type: ignore
     weather_sum = pd.DataFrame(
         {
-            "poa_global": 0,  # type: ignore
             "effective_irradiance": 0,  # type: ignore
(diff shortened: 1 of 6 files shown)


api/solarperformanceinsight_api/compute.py
```diff
@@ -254,9 +258,17 @@ def process_single_modelchain(
         )
     # mean
     weather_avg: pd.DataFrame = weather_sum / num_arrays  # type: ignore
-    adjusted_zenith: pd.DataFrame = adjust(
-        results.solar_position[["zenith"]]
-    )  # type: ignore
+    adjusted_zenith: pd.DataFrame
+    # not calculate if effective irradiance is provided
```
wholmgren, line 262:
```suggestion
    # not calculated if effective irradiance is provided
```

api/solarperformanceinsight_api/compute.py
```diff
@@ -254,9 +258,17 @@ def process_single_modelchain(
...
+    if results.solar_position is not None:
+        adjusted_zenith = adjust(results.solar_position[["zenith"]])  # type: ignore
+    else:
+        # calculate solar position making sure to shift times and shift back
+        adjusted_zenith = adjust(
+            mc.location.get_solarposition(
+                weather_avg.index.shift(freq=tshift)  # type: ignore
+            )[["zenith"]]
```
wholmgren, line 270: maybe worth noting modelchain passes pressure and air temperature to get_solarposition, but there's no difference if we extract zenith rather than apparent zenith.

https://github.com/pvlib/pvlib-python/blob/b666520f84cee895daa3418e06d8d8a543fc1b38/pvlib/modelchain.py#L1223-L1239

@cwhanse a nuance to be aware of

wholmgren (approved):

api/solarperformanceinsight_api/compute.py
cwhanse: @wholmgren are you saying that there's no practical difference between `zenith` and `apparent_zenith` in this use? Because there should be a difference in the values, albeit quite small.

wholmgren: no, I'm saying that because we use zenith we don't need to worry about inconsistencies between this explicit call to `get_solarposition` and the call within `ModelChain`. But that it would be a problem if we changed the SPI results to use apparent zenith rather than zenith.

PR #125, opened 2021-02-05
change gamma_pdc to units of %/C

### LINKED ISSUE #106: units of temperature coefficient
SPI currently uses temperature coefficients in units of per degree C, like pvlib.

In the Solar Forecast Arbiter project we changed to %/C based on stakeholder feedback. 

Should we do the same for SPI?

alorenzo175:
closes #106


api/solarperformanceinsight_api/models.py
```diff
@@ -90,6 +90,14 @@ def UserString(default: Any = Undefined, *, title: str = None, description: str
     )
 
 
+class PVLibBase(BaseModel):
+    """Provide a `pvlib_dict` method to convert parameters if needed
+    for using in pvlib"""
```
wholmgren, line 95:
```suggestion
    for using in pvlib. Child classes may implement model-specific conversions as needed."""
```

wholmgren (approved):

PR #130, opened 2021-02-09
Temperature param parsing at API

### LINKED ISSUE #129: Temperature parameter issue when creating pvwatts system
When creating a pvwatts system, the stored system ends up with default pvsyst temperature parameters. This results in empty temperature when the user clicks "edit" on the system.

lboeman:
closes #129 
Added a test that will serve as a regression test for this issue. Still unsure of cause.


alorenzo175:
Ah yeah the  problem is that pydantic can ignore extra parameters instead of forbidding/raising a validation error

alorenzo175:
I imagine with this though, maybe posting will fail validation?

lboeman:
> I imagine with this though, maybe posting certain things from the dashboard will fail validation?

I don't think so, If anything is invalid it should be fixed to match what the API expects anyway. Is there anything in particular you were thinking of?

alorenzo175:
> > I imagine with this though, maybe posting certain things from the dashboard will fail validation?
> 
> I don't think so, If anything is invalid it should be fixed to match what the API expects anyway. Is there anything in particular you were thinking of?

Nothing in particular, but I thought maybe the invalid spec from #129 originated on the dashboard. I can't reproduce any issue like that now though

lboeman:
> > > I imagine with this though, maybe posting certain things from the dashboard will fail validation?
> > 
> > 
> > I don't think so, If anything is invalid it should be fixed to match what the API expects anyway. Is there anything in particular you were thinking of?
> 
> Nothing in particular, but I thought maybe the invalid spec from #129 originated on the dashboard. I can't reproduce any issue like that now though

Gotcha. Yes, the issue was that when saving a valid PVWatts system, the api would save a System with the default PVSystTemperatureParameters.

Can be reproduced on the dev dashboard by creating a PVWatts system, then when you hit edit the temperature parameters will be empty because the dashboard assumes it will be SAPM Temperature Parameters and can't access the correct properties.

Thanks for helping figure it out, I was thrown that temperature parameters were the only place this was an issue. OK to merge?

alorenzo175 (approved):

Issue #115, opened 2021-02-02
Combine all job results into a single download


alorenzo175:
Probably as a ZIP file otherwise we have to worry about column names and possible index mismatch. Could implement in API or in JS

PR #136, opened 2021-02-16
User configurable plots

### LINKED ISSUE #126: Add ability to plot separate result timeseries as traces on same plot
Example was that it would be nice to plot performance against ghi for a calculate workflow.

Alternative would be to generate a plot for each timeseries.

Might be interesting to implement both ideas, allowing a user to add multiple plots to the page with whatever traces they'd like to plot. 

If allowing users to create plots of whatever timeseries they would like they should also be able to download a csv of the plotted data.

lboeman:
closes #126


lboeman:
Adds the user created configurable plots that were demo'd on Monday's call. Adding multiple plots is now working, and I've added friendlier table headers with units to the tables as shown in the screenshots below.
Screenshot from 2021-02-17 10-21-28
Screenshot from 2021-02-17 10-21-37
We talked about merging the columns of these two tables and I have a couple of questions about that.
- The "Monthly Summary" table contains summary data for the expected performance should any fields be prefixed with "Expected" when merging tables? This would go for "Reference" as well, when other data is available for 2A and 2B.
- Are the columns of interest from the monthly summary table POA insolation, Effective Insolation, and Average Daytime Cell Temperature?

In addition, what are the feelings on using these configurable plots on all results pages? I had originally though about keeping the simple single plot option for the calculation workflows but I think this might work for everything.

ncroft-b4:
> The "Monthly Summary" table contains summary data generated when computing the expected performance. Should any fields be prefixed with "Expected" when merging tables? This would go for "Reference" as well, when other data is available for 2A and 2B.

I don't think that is necessary. The current labels are good, I think. Rather than "Expected" the other fields like POA and Temperature are "Measured" (or maybe "Calculated" if GHI was measured - being technically accurate will get tricky), and "Expected" would imply something else and cause confusion. However, when moving to 2A and 2B, I do think "Reference" will be an important thing to include for the uploaded reference poa and temperature data. In fact, that makes me realize we should discuss what to include in the output of those workflows pretty soon, as we should probably either commit to include both the reference and the measured weather data in the table, or neither. 

> Are the columns of interest from the monthly summary table POA insolation, Effective Insolation, and Average Daytime Cell Temperature?

Just POA and Cell Temp, I suggest leaving off Effective Insolation. That should be included in the options of downloadable data for user who are interested in it, but not necessary for the table I don't think. That would put us at 6 data columns in the output table and that seems good. Also, it's minor, but if eventually we could round off the decimals of the Wh and C data and round the Ratio to two or three digits, or even display that as a percentage to one digit, that would be my suggestion.

lboeman:
@ncroft-b4 thanks for the feedback and clarification.

Here's a screenshot of the updates to the results page:

Screenshot_2021-02-22 solarperformanceinsight(1)

Summarizing the updates: 
- I've added a table of the results files for users to download, and to provide an overview of all the available data. There's no direct way to download the data in the Monthly Summary table, but all of that data is available between the `monthly summary` and `actual vs expected energy` result files. If a single file containing all of the monthly data is desired perhaps we can work it into the PR for #115.
- Added a button for updating the title of the plots in the "Timeseries Results" section. The widget for updating the title can be seen in the box for the 
- Rounded the values in the summary table and displayed the "Ratio" field as a percentage rounded to one decimal. This rounding is only happening in the display code, and data downloaded by the user will contain the original, un-rounded values. The downloaded value should have a reasonable amount of precision, without the floating point errors that cause the large numbers on the dashboard. 

@alorenzo175 @wholmgren @cwhanse Any more thoughts on this PR?

wholmgren:
looks great to me. I think we should merge it sooner than later so that others can play with it while you move on to other things.

lboeman:
Alright, going to merge so we can get this in dev for people to use. Happy to address feedback in issues if there is any.

PR #137, opened 2021-02-17
change default aoi model to physical

### LINKED ISSUE #134: change default AOI model to physical
It appears that the AOI model passed to `ModelChain` is `'no_loss'`:

https://github.com/SolarPerformanceInsight/solarperformanceinsight/blob/b83658605d4a32dfd5ef8c78533c35252f38f30c/api/solarperformanceinsight_api/models.py#L486

and that there is no way for a user to change it at this time. I think we should change the default to `'physical'` (`AOIModelEnum.physical`) and make a new issue for user input.

wholmgren:
closes #134


wholmgren:
@lboeman can I turn this over to you for the database stuff? Just push directly to this branch.

alorenzo175:
Should be able to just change the aoi_model in the example system at the top of models.py to "physical".
In the database, I don't think aoi_model is explicitly set, so it will use whatever is the default

wholmgren:
Thanks, that worked. I thought I was going to have to wade into sql stuff I don't understand.

alorenzo175 (approved):

PR #141, opened 2021-02-18
Use case 2A model changes

alorenzo175:
Update model.py to support use case 2A


alorenzo175:
@lboeman what do you think of the model changes? Basically, what was in `Job.parameters.job_type` is at the `Job.parameters` level. The `ComparePredictedActualJobParameters` will need special handling, but at least I think the model for UC 2B will be somewhat similar. Any other ideas?

lboeman:
> @lboeman what do you think of the model changes? Basically, what was in `Job.parameters.job_type` is at the `Job.parameters` level. The `ComparePredictedActualJobParameters` will need special handling, but at least I think the model for UC 2B will be somewhat similar. Any other ideas?

I think the changes are generally good and the pairing of granularity/data availability properties makes sense. I had some lofty ideas about combining classes into the original job parameters using the `*_data_parameters` for predicted, expected, and actual for all job types but after working through this a few times I realized that just complicates things. I like where this is at. 

It seems like handling of the monthly flag could be its own class. At least for the purposes of demonstrating the restrictions in the API docs. e.g. requires ac and weather, granularity of "system" and such. That way instead of responding with a "you must designate ac and weather when using monthly data" when a user tries to create a job with an invalid `data_available` the API just knows to create data objects with the correct expected columns

alorenzo175:
> It seems like handling of the monthly flag could be its own class. At least for the purposes of demonstrating the restrictions in the API docs. e.g. requires ac and weather, granularity of "system" and such. That way instead of responding with a "you must designate ac and weather when using monthly data" when a user tries to create a job with an invalid `data_available` in their `predicted_data_parameters` the API just knows to create data objects with the correct expected columns

Good idea

alorenzo175:
@lboeman are the adjustments to the dashboard for the new schemas something I can leave to you?

lboeman:
> @lboeman are the adjustments to the dashboard for the new schemas something I can leave to you?

Yep.

alorenzo175:
cool. We can hold off on merging this until the dashhboard is updated, or just wait to deploy

lboeman (approved):

PR #142, opened 2021-02-18
implement CEC (SAM) module model

cwhanse:
Adds `CECModuleParameters` class to `model.py`


wholmgren:
@cwhanse what temperature model should go with this new workflow that uses the CEC modules? For SPI's "pvwatts" workflow, we use the SAPM temperature model. For SPI's "pvsyst" workflow, we use the pvsyst temperature model.

cwhanse:
> @cwhanse what temperature model should go with this new workflow that uses the CEC modules? For SPI's "pvwatts" workflow, we use the SAPM temperature model. For SPI's "pvsyst" workflow, we use the pvsyst temperature model.

Looks like the default is a version of the NOCT cell temperature model which isn't in pvlib (yet): page 58 [here](https://www.nrel.gov/docs/fy18osti/67399.pdf).

For now, wire it to the pvsyst temperature model. I'll see about a PR to pvlib to add the SAM default.

wholmgren:
My reading of the pvlib fuentes issues/prs is that we should use that model both pvwatts and sam workflows, assuming we're ok with the slow performance.

cwhanse:
> My reading of the pvlib fuentes issues/prs is that we should use that model both pvwatts and sam workflows, assuming we're ok with the slow performance.

True for PVWatts. SAM offers several "model chains", one being PVWatts (using Fuentes as the temperature model), another being the CEC module model which uses the NOCT cell temperature model as the default (Fuentes is exposed here as an option). At least that's the way I understand SAM. @kanderso-nrel?

kanderso-nrel:
PVWatts uses Fuentes.  Side note: to get around the slow performance of the pvlib implementation, I've been using my own numba version until I finally figure out how to finish the more general pvlib numba PR.  Happy to send that code around if it would be useful.

For the CEC "model chain": I'm not an expert here, but from what I see in the SAM (v2020.11.29) docs, Fuentes is not used for CEC.  The alternative to the NOCT model is apparently a model by Neises that I'm not familiar with.  Here's a screenshot of SAM's help (Detailed PV model > module):

<details>
  <summary>Click to expand!</summary>

image

</details>

Link to the SAM page where the Neises PDF can be downloaded: https://sam.nrel.gov/photovoltaic/pv-sub-page-2.html

lboeman:
Added the dashboard side of things here. The CECModuleParameters are available in a new "SAM" model option that is otherwise the same as the PVSyst-like option. The CEC Module database is available, and I've added the fuzzy search feature. The  search should now return all values that contain all of the words in the search phrase, without regard for case. e.g. Users could get close to the module "Alps_Technology_ATI_M660_240" by searching something like "M660 ati alps".

Not sure what happened here with the diff, it's displaying the changed files from the last few pull requests that were merged. I'll try to get that worked out.

lboeman:
Merged #149, so closing this PR.

PR #145, opened 2021-02-23
Allow users to define performance units

### LINKED ISSUE #98: Add Power units to upload mapping process
Allow users to define what units they have in their data files. This will allow us to nail down expected units at the API for simpler code there. Conversion to expected base units before API submission. Users should also be able to select units on the results page as well, so standalone conversion functions are a good idea.

lboeman:
closes #98

### DIFF
--- a/dashboard/src/components/jobs/CSVMapper.vue
+++ b/dashboard/src/components/jobs/CSVMapper.vue
@@ -187,7 +187,7 @@ export default class CSVMapper extends Vue {
     // pop the index from the mapping
     const loc = newMap.loc;
     newMap = { ...newMap };
-    newMap["time"] = { csvHeader: this.timeField };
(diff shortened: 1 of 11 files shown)


lboeman:
Adds an extra entry when users are mapping performance uploads to define units. This adjusts the format of mappings from a string to string mapping to a string to object mapping where the object has a `csv_header` field and a  an optional `units` field.
Screenshot from 2021-02-24 10-58-45

alorenzo175 (approved):
We'll probably need to add Wh conversions for energy uploads and for display conversions of the results table, and have to ask how common irradiance in kW/m^2 is, but should be easy to add those later

alorenzo175:
And I can't see a separate issue for units conversion of results as displayed on teh dashboard, so might be good to add an issue to track that

PR #147, opened 2021-02-24
Improve job time parameter interface

### LINKED ISSUE #113: Add helper tooltips to the job kickoff pages
at least for the time parameters so it's more clear what the acceptable formats are.

lboeman:
closes #113

Updates the time parameters with a datepicker.
Screenshot from 2021-02-24 14-47-25
This datepicker is backed by luxon, and has a few other features that make validating the time parameters simpler. The timezone selected by the user is used for the timezone of the start/end time in real time. This is why I've moved the timezone select about the start/end; the start/end adjust when the timezone changes. We also don't have to rely on users typing in a date in a specific format.


dashboard/src/components/jobs/parameters/TimeParameters.vue
```diff
@@ -42,11 +69,19 @@ export default class JobTimeParameters extends Vue {
...
+      .setZone(zone)
+      .set({
+        second: 0,
+        milliseconds: 0
+      });
     return {
-      start: "2020-01-01T00:00+00:00",
-      end: "2020-02-01T00:00+00:00",
+      start: now.toISO(),
+      end: now.toISO(),
```
alorenzo175, line 82: Do we need to set these to not null initially? One issue is that the start picker max time is restricted to end, so in a normal workflow you wouldn't be able to pick a start in the future (and end after start). No reason we should restrict the range if not necessary

lboeman: I've updated these to initialize to null and the "Get Started" button is disabled until the user makes a selection for start and end.

alorenzo175 (approved):

PR #149, opened 2021-02-25
implement CEC (SAM) module model Copy

lboeman:
Opening a pull request from my own fork to fix the wonky diff I created in #142


api/solarperformanceinsight_api/models.py
```diff
@@ -227,6 +227,74 @@ def pvlib_dict(self):
...
+    )
+    dEgdT: float = Field(
+        -0.0002677,
+        description=(
+            "The temperature dependence of the energy bandgap at reference "
+            "conditions in units of 1/K. -0.0002677 1/K for all modules in "
+            "the CEC database."
+        ),
+    )
+    _modelchain_dc_model: str = PrivateAttr("pvsyst")
```
alorenzo175, line 295:
```suggestion
    _modelchain_dc_model: str = PrivateAttr("cec")
```

api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -17,6 +17,11 @@
     .astype(float)
 )
 
+# Database does not provide the fields 'cells_in_series', 'EgRef', 'dEgdT'
+# which have defaults in the model.
+cec_keys = ["alpha_sc", "a_ref", "I_L_ref", "I_o_ref", "R_sh_ref", "R_s", "Adjust"]
```
alorenzo175, line 22:
```suggestion
cec_keys = ["alpha_sc", "a_ref", "I_L_ref", "I_o_ref", "R_sh_ref", "R_s", "Adjust", "N_s"]
```

api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -17,6 +17,11 @@
...
+cec_module_params = retrieve_sam("CECMod").loc[cec_keys].astype(float)
```
alorenzo175, line 23:
```suggestion
cec_module_params = retrieve_sam("CECMod").loc[cec_keys].astype(float).rename(index={"N_s": "cells_in_series"})
```

api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -50,3 +55,37 @@ def get_sandia_inverter(
...
+    )
+) -> models.CECModuleParameters:
+    """Get the parameters for the named Sandia inverter"""
+    if module_name not in cec_module_params.columns:
+        raise HTTPException(
+            status_code=404,
+            detail=f"Module parameters for {module_name} not found",
+        )
+    return models.CECModuleParameters(
+        **cec_module_params[module_name].to_dict(), cells_in_series=0
```
alorenzo175, line 90:
```suggestion
        **cec_module_params[module_name].to_dict()
```

dashboard/src/views/SystemSpec.vue
```diff
@@ -111,7 +114,11 @@ export default class SystemSpec extends Vue {
   data() {
     return {
       system: this.system,
-      modelPresetOptions: ["pvsyst", "pvwatts"],
+      modelPresetOptions: {
+        pvsyst: "pvsyst-like",
+        pvwatts: "pvwatts",
+        sam: "SAM"
+      },
       model: "pvsyst",
```
alorenzo175, line 122: I think we probably want to set SAM as the default since it provides searchable modules + inverters for new users.

alorenzo175: Thoughts @wholmgren @cwhanse ?

api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -50,3 +55,37 @@ def get_sandia_inverter(
...
+    "/cecmoduleparameters/{module_name}",
+    response_model=models.CECModuleParameters,
+    responses={200: {}, 404: {}, 422: {}},
+    summary="Get a set of CECModuleParameters",
+)
+def get_cec_module(
+    module_name: str = Path(
+        ...,
+        description="Name of the module to fetch parameters for",
+        example="A10Green Technology A10J-S72-175",
```
alorenzo175, line 80:
```suggestion
        example="Canadian_Solar_Inc__CS5P_220M",
```
Should also make the app test cover lns 89-90

dashboard/src/components/Browser.vue
```diff
@@ -1,8 +1,6 @@
 <template>
   <div class="db-browser">
-    <input v-model="search" v-on:keyup.enter="filterOptions" />
-    <button class="search" @click="filterOptions">Search</button>
-    <button class="search-reset" @click="resetSearch">Reset Search</button>
+    <input v-model="search" class="search-box" v-on:keyup="filterOptions" />
```
alorenzo175, line 3: any way to make a double click on an item select it vs scrolling to the bottom?

alorenzo175: also not sure about keyup vs enter/click for searching modules; seems to bog my circa 2012 i5 and browser down

alorenzo175 (commented):
good start. seems like there are a few places that have tests for the other models but not the new cec

api/solarperformanceinsight_api/routers/parameters.py
lboeman: Is this just a more common module?

dashboard/src/components/Browser.vue
lboeman: Cool, thanks for checking it out. I've reverted this to the enter/click behavior since it seems like the set/clear timeout is not behaving as expected.

api/solarperformanceinsight_api/routers/parameters.py
alorenzo175: It's roughly the one in the pvlib tutorial, and the previous example was not a valid key so returned 400

api/solarperformanceinsight_api/routers/parameters.py
```diff
@@ -17,6 +17,25 @@
...
+    "I_o_ref",
+    "R_sh_ref",
+    "R_s",
+    "Adjust",
+    "N_s",
+]
+cec_module_params = (
+    retrieve_sam("CECMod")
+    .loc[cec_keys]
+    .astype(float)
```
alorenzo175, line 35: I guess json/javascript only have "number" and don't differentiate between int and float, right? just because now cells in series is an int and should roundtrip back as the same int

lboeman: Yes, javascript only has the `number` type, but JSON can represent integers correctly. The model is deserializing the `cells_in_series` field to the correct int before transmitting it.
On the dashboard, this is being handled by checking the schema for the `integer` type and setting the step of the input to 1, so that an integer is included in the JSON when posting.

alorenzo175 (approved):

PR #153, opened 2021-02-26
Support uploading original/predicted data with a different year for UC2A

### LINKED ISSUE #156: Handle DST transitions when localizing
Sentry Issue: [API-3A](https://sentry.io/organizations/solar-performance-insight/issues/2239890535/?referrer=github_integration)

```
NonExistentTimeError: 1990-04-01 02:00:00
(12 additional frame(s) were not displayed)
...
  File "pandas/core/generic.py", line 9648, in tz_localize
    ax = _tz_localize(ax, tz, ambiguous, nonexistent)
  File "pandas/core/generic.py", line 9636, in _tz_localize
    ax = ax.tz_localize(tz, ambiguous=ambiguous, nonexistent=nonexistent)
  File "pandas/core/indexes/datetimes.py", line 235, in tz_localize
    arr = self._data.tz_localize(tz, ambiguous, nonexistent)
  File "pandas/core/arrays/datetimes.py", line 970, in tz_localize
    new_dates = tzconversion.tz_localize_to_utc(
  File "pandas/_libs/tslibs/tzconversion.pyx", line 319, in pandas._libs.tslibs.tzconversion.tz_localize_to_utc
```

alorenzo175:
from the specified job time parameters. Only looks at the first
timestamp of the data and uses it to shift the entire frame. 
Removes feb 29 from shifted data if present

closes #156


alorenzo175:
@wholmgren are you ok with not doing anything about leap years? As in if the reference data is from a leap year, it may have that one extra day that is ignored at the end of the series. If we're adjusting to a leap year from a non leap year (and have a  year of data) 12/31 will be missing.

wholmgren:
> are you ok with not doing anything about leap years?

I'm ok with it, but I would like to do more if it's simple. I anticipated that it would require only one or two `if calendar.isleap` blocks with fairly straightforward code. Maybe a follow up issue would be ok too.

> As in if the reference data is from a leap year, it may have that one extra day that is ignored at the end of the series.

So Dec 31 in the reference data would be ignored? I would have thought Feb 29 would be ignored. For annual comparisons I don't think it matters, but for month to month comparisons I think it would make more sense to users if February was the odd month instead of December. 

> If we're adjusting to a leap year from a non leap year (and have a year of data) 12/31 will be missing.

On the call we discussed doing something simple like duplicating input data from Feb 28 into Feb 29 or something a little more complicated like averaging results from Feb 28 and Mar 1 into Feb 29.

cwhanse:
>  I would have thought Feb 29 would be ignored. 

I agree.

> On the call we discussed doing something simple like duplicating input data from Feb 28 into Feb 29 or something a little more complicated like averaging results from Feb 28 and Mar 1 into Feb 29.

I think simply dropping Feb 29 from measured (or reference) is easier both to implement and explain.

alorenzo175:
>    I'm ok with it, but I would like to do more if it's simple. I anticipated that it would require only one or two if calendar.isleap blocks with fairly straightforward code. Maybe a follow up issue would be ok too.

It isn't straightforward once you take into account that the job start time could be any arbitrary time and need not extend for a year; at least from my perspective. 

A consequence of this is that the data may not be shifted properly if, for example, job start = 2020-12-31T17:00-07:00 (end at start + 1year) and reference data starts at 1990-01-01T00:00:00-07:00. Here the reference data would be shifted to 2020, but the majority of the actual data is in 2021. The upload will fail at least. We could look at the year mode for reference and job index to find the shift, but I bet there are still issues with that.

Looks like DST transitions are another kink

> I think simply dropping Feb 29 from measured (or reference) is easier both to implement and explain.

Agreed and what is now implemented

api/solarperformanceinsight_api/utils.py
```diff
@@ -172,32 +172,43 @@ def standardize_months(df: pd.DataFrame) -> pd.DataFrame:
...
-                newdf.index
-            ).to_pydatetime()
+            jobtimeindex._time_range.difference(index).to_pydatetime()  # type: ignore
         )
     else:
         extra = []
         missing = []
+    newdf.index = index
+    # drop possible duplicate feb 28 when going from leap year to non
+    # and possible duplicate times from DST transitions
```
alorenzo175, line 210: A way around losing data is probably by converting everything to UTC, using that for calculations, then converting back to aggregate/display. Right now that would take too long compared to dropping a few data points at night

alorenzo175:
closes #156

alorenzo175:
I've standardized on `ambiguous=True, nonexistent="shift_forward"` for any timezone localizing around DST issues. Thoughts @wholmgren?

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.tz_localize.html

wholmgren:
I don't understand `ambiguous=True`. I thought you needed to supply an array of bools. Seems to me that pandas should raise an error if you try to set True to a timestamp for which it's not possible. In any case, why doesn't `ambiguous='infer'` work for our use case? Maybe `NaT` if infer fails?

`nonexistent="shift_forward"` is probably ok, though I might have gone with `NaT`.

lboeman (approved):
Looks good to me.

PR #157, opened 2021-03-01
Usecase 2A Monthly Job setup

lboeman:
Updates for monthly job setup and data upload for 2A for the dashboard.

The fork in workflow occurs when the user selects the "predicted performance to actual performance" option in the comparison menu. The user is presented with a "What is the time resolution of your data?". If they select the monthly option they are prompted with the expected variables. 
Screenshot from 2021-03-01 11-21-25


dashboard/src/components/jobs/JobHandler.vue
```diff
@@ -156,7 +185,22 @@ Component that handles basic job/workflows.
               :irradiance_type="jobParameters.irradiance_type"
               :data_objects="filteredDataObjects(step)"
             >
-              <b>Upload Predicted Performance</b>
+              <b>upload predicted performance</b>
```
alorenzo175, line 188: Why the case change?  here and below

dashboard/src/components/jobs/parameters/JobParams.vue
```diff
@@ -188,19 +201,28 @@ export default class JobParameters extends Vue {
...
   get jobSpec() {
-    return {
-      system_id: this.systemId,
-      time_parameters: this.timeParams,
-      weather_granularity: this.weather_granularity,
-      irradiance_type: this.irradiance_type,
-      temperature_type: this.temperature_type,
-      ...this.jobTypeParams
-    };
+    if (this.isMonthly) {
```
alorenzo175, line 204: A few additional tests to cover this case and a few of the other easy to test changes would be good

dashboard/src/components/jobs/parameters/CompareJobParams.vue
```diff
@@ -43,7 +42,27 @@
...
         </div>
-        <div class="my-1">
+        <div v-if="containsPredicted">
+          <p>What is the time resolution of your data?</p>
+          <div class="ml-1 mt-1">
+            <input
+              type="radio"
+              id="hourly-resolution"
+              v-model="timeResolution"
+              value="leHourly"
```
alorenzo175, line 52: I wouldn't mind leaving this greyed and disabled since form changes are still required

dashboard/src/components/jobs/JobHandler.vue
lboeman: Thanks for catching that. I don't know what did this. I think it slipped in with some linting and broke some of the components too.

dashboard/src/components/jobs/parameters/CompareJobParams.vue
```diff
@@ -43,7 +42,30 @@
...
+            <label for="hourly-resolution">My data is hourly or better.</label>
+            <br />
+            <input
+              disabled="true"
+              type="radio"
+              id="monthly-resolution"
+              v-model="timeResolution"
+              value="monthly"
+            />
+            <label class="greyed" for="monthly-resolution">
```
alorenzo175, line 63: whoops, I meant grey the hourly. I think monthly will be available sooner

alorenzo175 (approved):

dashboard/src/components/jobs/parameters/CompareJobParams.vue
lboeman: Gotcha. duh, that makes more sense.

PR #162, opened 2021-03-04
Improve timefield

### LINKED ISSUE #154: improve date/time picker
The date picker looks nice and works well for recent dates, but it is extremely tedious to specify a start and end date like jan 1, 2019 through dec 31, 2019. 

It's also not clear that the time picker will come after the date picker, so I was surprised as a first time user. Is it possible to set the default start time to 00:00 and the default end time to 23:59?

Maybe a menu based system would be easier. Some users could still type the date if they tab between fields and take advantage of browser autoselection, while users could still point and click.

lboeman:
closes #154

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -15,7 +15,6 @@
     "@sentry/browser": "^5.27.6",
     "@sentry/integrations": "^5.27.6",
     "@types/papaparse": "^5.2.4",
-    "@types/vue-datetime": "^1.0.1",
     "ajv": "^7.0.0-beta.6",
     "apache-arrow": "^3.0.0",
     "core-js": "^3.6.5",
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -24,7 +24,6 @@
     "plotly.js-basic-dist": "^1.58.4",
     "vue": "^2.6.11",
     "vue-class-component": "^7.2.3",
-    "vue-datetime": "^1.0.0-beta.14",
     "vue-property-decorator": "^8.4.2",
     "vue-router": "^3.2.0",
     "vue-template-loader": "^1.1.0",
--- a/dashboard/src/components/jobs/parameters/TimeParameters.vue
+++ b/dashboard/src/components/jobs/parameters/TimeParameters.vue
@@ -30,14 +30,18 @@
     </div>
     <div class="timefield">
       <b>Start:</b>
-      <datetimefield @update-datetime="setStart" :timezone="timezone" />
-      <help helpText="The value of the first timestamp in your data." />
+      <datetimefield
+        @update-datetime="setStart"
+        :timezone="timezone"
+        helpText="The value of the first timestamp in your data."
+      />
     </div>
     <div class="timefield">
       <b>End:</b>
-      <datetimefield @update-datetime="setEnd" :timezone="timezone" />
-      <help
-        helpText="The last minute covered by your data. For example, 60 minute data with a last timestamp at 2020-12-31 23:00 should have an end of 2020-12-31 23:59."
+      <datetimefield
+        @update-datetime="setEnd"
+        :timezone="timezone"
+        helpText="The end date and time of your data. This may be any between the last timestamp and the beginning of the next timestep. For example, 60 minute data with a last timestamp at 2020-12-31 23:00 should have an end between 2020-12-31 23:01 and 2021-01-01 00:00."
       />
     </div>
     <div v-if="errors">
--- a/dashboard/src/components/jobs/parameters/TimeParameters.vue
+++ b/dashboard/src/components/jobs/parameters/TimeParameters.vue
@@ -13,7 +13,7 @@
         <option v-for="tz in timezoneList" :key="tz">{{ tz }}</option>
       </select>
       <help
-        helpText="The timezone used in your data. If your data does not account for daylight savings, use the appropriate fixed-offset. (e.g. Etc/GMT-7)"
+        helpText="The timezone used in your data. If your data does not account for daylight savings, use the appropriate fixed-offset. (e.g. Etc/GMT+7)"
       />
     </div>
     <div class="timefield">
--- a/dashboard/src/components/jobs/parameters/TimeParameters.vue
+++ b/dashboard/src/components/jobs/parameters/TimeParameters.vue
@@ -54,7 +54,7 @@
   </div>
 </template>
 <script lang="ts">
-import { Component, Vue } from "vue-property-decorator";
+import { Component, Vue, Prop } from "vue-property-decorator";
 import Timezones from "@/constants/timezones.json";
 import { LocalZone } from "luxon";
 import { Datetime as DatePicker } from "vue-datetime";
@@ -64,8 +64,9 @@ Vue.component("datetime", DatePicker);
 
 @Component
 export default class JobTimeParameters extends Vue {
-  start!: string;
-  end!: string;
+  @Prop() timeparams!: Record<string, any>;
+  start!: string | null;
+  end!: string | null;
   timezone!: string;
   step!: number;
   timezoneList: Array<string> = Timezones;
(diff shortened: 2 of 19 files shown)


lboeman:
Created a new datetime input field. Each component of the date time (year, month, day, hour, minute) has it's own input which allows for quickly tabbing through the options. When all of the fields for one of the datetimes is filled in, a luxon Datetime object is instantiated and any errors are reported (an example error is shown below).  I've also added some explanatory text for the timezone and start/end fields, hopefully this clarifies how users should set those fields.

Screenshot from 2021-03-05 16-35-12

Also created a component for displaying API errors that can be used elsewhere. One gotcha I discovered here is that the API responds with errors for every possible class in a union when a failure occurs, so relying on the API errors here results in repeated errors (one for every job parameter type). So I've just added a check for start < end before allowing a post.

dashboard/src/components/jobs/parameters/TimeParameters.vue
```diff
@@ -12,6 +12,9 @@
       >
         <option v-for="tz in timezoneList" :key="tz">{{ tz }}</option>
       </select>
+      <help
+        helpText="The timezone used in your data. If your data does not account for daylight savings, use the appropriate fixed-offset. (e.g. Etc/GMT-7)"
```
alorenzo175, line 16:
```suggestion
        helpText="The timezone used in your data. If your data does not account for daylight savings, use the appropriate fixed-offset. (e.g. Etc/GMT+7)"
```

I don't know the exact reasoning but +7 is the right zone for Arizona

dashboard/package.json
```diff
@@ -24,7 +24,6 @@
     "plotly.js-basic-dist": "^1.58.4",
     "vue": "^2.6.11",
     "vue-class-component": "^7.2.3",
-    "vue-datetime": "^1.0.0-beta.14",
```
alorenzo175, line 27: don't need it's @types either right?

dashboard/src/components/jobs/parameters/TimeParameters.vue
```diff
@@ -27,40 +30,33 @@
...
-        placeholder="Click here to set start date"
-        v-model="end"
-        :zone="timezone"
-        :picker-zone="timezone"
-        :min-datetime="start"
-        format="y-LL-dd HH:mmZZ"
-      ></datetime>
+      <datetimefield @update-datetime="setEnd" :timezone="timezone" />
+      <help
+        helpText="The last minute covered by your data. For example, 60 minute data with a last timestamp at 2020-12-31 23:00 should have an end of 2020-12-31 23:59."
```
alorenzo175, line 40: This is actually a little more flexible. In this example anything after 23:00 would work and 2021-01-01 00:00 would also work. Not sure of a good way to describe it though

dashboard/src/components/jobs/parameters/TimeParameters.vue
```diff
@@ -27,40 +30,33 @@
...
+      />
+    </div>
+    <div v-if="errors">
```
alorenzo175, line 43: can we move these out a div level (I think) so they come on a line after End: instead of between the field and tooltip box

alorenzo175 (commented):
This is much better than the picker. Something that would be nice is if the display would keep leading zeros 
image

lboeman:
> This is much better than the picker. Something that would be nice is if the display would keep leading zeros
> image

I agree, I spent a little time looking into this and it doesn't seem like a simple fix unfortunately.

dashboard/src/components/jobs/parameters/TimeParameters.vue
```diff
@@ -27,40 +30,37 @@
...
-        v-model="end"
-        :zone="timezone"
-        :picker-zone="timezone"
-        :min-datetime="start"
-        format="y-LL-dd HH:mmZZ"
-      ></datetime>
+      <datetimefield
+        @update-datetime="setEnd"
+        :timezone="timezone"
+        helpText="The end date and time of your data. This may be any between the last timestamp and the beginning of the next timestep. For example, 60 minute data with a last timestamp at 2020-12-31 23:00 should have an end between 2020-12-31 23:01 and 2021-01-01 00:00."
```
lboeman, line 44: @alorenzo175 What do you think of this end time explanation?

alorenzo175:
> > This is much better than the picker. Something that would be nice is if the display would keep leading zeros
> > image
> 
> I agree, I spent a little time looking into this and it doesn't seem like a simple fix unfortunately.

What about a function like 
```javascript
leadingZero(what: string, event){
    this[what] = parseInt(event.target.value);
    const padded = this[what].toString().padStart(2, "0");
    event.target.value = padded;
}
```

and the input for e.g. minute would be 
`<input class="minute" ... @change="leadingZero('minute', $event)"/>` with no v-model

dashboard/src/components/jobs/parameters/TimeParameters.vue
alorenzo175: I think something a little simpler like 
```suggestion
        helpText="The end date and time of your data, exclusive. For example, 60 minute data with a last timestamp at 2020-12-31 23:00 should have an end between 2020-12-31 23:01 and 2021-01-01 00:00."
```

lboeman:
> > > This is much better than the picker. Something that would be nice is if the display would keep leading zeros
> > > image
> > 
> > 
> > I agree, I spent a little time looking into this and it doesn't seem like a simple fix unfortunately.
> 
> What about a function like
> 
> ```js
> leadingZero(what: string, event){
>     this[what] = parseInt(event.target.value);
>     const padded = this[what].toString().padStart(2, "0");
>     event.target.value = padded;
> }
> ```
> 
> and the input for e.g. minute would be
> `<input class="minute" ... @change="leadingZero('minute', $event)"/>` with no v-model
This does not play nice with browsers that enforce the `type="number"` attribute like Chrome. We could use a regular text input and do all of the checking in javascript, but relying on the browsers immediate feedback is nice.

dashboard/src/components/jobs/parameters/TimeParameters.vue
cwhanse: Working at that a bit more:

""The end date and time of your data. End is the timestamp at the beginning of the last interval. For example, 60 minute data with the last interval covering the period 2020-12-31 23:00 to 2020-12-31 23:59:59, End should have the timestamp at 2020-12-31 23:00."

lboeman: I should have left this in a single comment thread. Per Tony's previous comment:

> This is actually a little more flexible. In this example anything after 23:00 would work and 2021-01-01 00:00 would also work. Not sure of a good way to describe it though

As I understand it, the end date is acting as an exclusive bound on the timestamps and not necessarily the timespan covered by the data. I think this is primarily because these time parameters are used to generate an expected index for validating the data and that's just Pandas behavior?

alorenzo175:
Ok, last try. On chrome and firefox on fedora for me, just getting using a standard `v-model=` works in keeping the zero if typed. Does that usually work and not cause type issues?

dashboard/src/components/jobs/parameters/TimeParameters.vue
alorenzo175: > As I understand it, the end date is acting as an exclusive bound on the timestamps and not necessarily the timespan covered by the data. I think this is primarily because these time parameters are used to generate an expected index for validating the data and that's just Pandas behavior?

Yes it is an exclusive bound. We decided that data would be assumed to be, in arbiter speak, beginning interval labelled, and it is easier to specify a time range like 2020-01-01T00:00 to 2021-01-01T00:00 than set end = 2020-12-31T23:59. Pandas is happy to include both endpoints if they fit properly in the specified frequency so there is a check to drop the last index point if it == end.

lboeman: Alright, so I've included @alorenzo175's suggestion here, which reflects the current behavior. If we need to adjust the way the end parameter is handled I think we should open an issue for that.

lboeman:
> Ok, last try. On chrome and firefox on fedora for me, just getting using a standard `v-model=` works in keeping the zero if typed. Does that usually work and not cause type issues.

This is working now. Looks like luxon can handle the 0-padded values and validates them appropriately. Thanks for taking a look at that. 

Everything else look good?

alorenzo175 (approved):

PR #165, opened 2021-03-08
Add CEC gamma_r parameter

alorenzo175:
The temperature coefficient of power will be required for 2A use
cases. PVWatts systems already have this defined as gamma_pdc, gamma_r
is the equivalent for CEC. PVSyst will need some kind of conversion
from the single diode parameters later.


api/solarperformanceinsight_api/models.py
```diff
@@ -266,6 +266,13 @@ class CECModuleParameters(PVLibBase):
     R_s: float = Field(
         ..., description="Series resistance at reference conditions, in ohms"
     )
+    gamma_r: float = Field(
+        ...,
+        description=(
+            "Temperature coefficient of power in units of %/C. "
+            "Typically -0.2 to -0.5 % per degree C"
```
alorenzo175, line 273: @cwhanse is this an appropriate description?

lboeman (approved):

PR #169, opened 2021-03-10
Add compute functions for use case 2A

alorenzo175:
In order to convert from any temperature and irradiance input to cell temp and POA global, this just runs the full pvlib.ModelChain on the the reference and actual weather and grabs the results. It also calculates the adjusted power at the inverter level and sums up to the system level. POArat and TempFactor are calculated at the array level (with the array gamma_pdc, poas, and temps) and the simple mean is taken to get to the inverter level.


api/solarperformanceinsight_api/compute.py
```diff
@@ -444,3 +501,182 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
...
+            chain.results.total_irrad, chain.results.effective_irradiance
+        )
+        poa_actual = _irrad_for_celltemp(
+            chain_actual.results.total_irrad, chain_actual.results.effective_irradiance
+        )
+        # use cell temperature from the pvlib modelchain
+        # If air temp + wind speed or module temperature were supplied, they are
+        # converted to cell temperature via the array's temperature model
+        t_ref = chain.results.cell_temperature
+        t_actual = chain_actual.results.cell_temperature
```
alorenzo175, line 624: test this specifically when providing module_temperature for the two temp models

api/solarperformanceinsight_api/compute.py
```diff
@@ -444,3 +501,182 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
...
+
+        # mean of array POArat * TempFactor for this inverter
+        # could make more sense to use weighted mean with weights set
+        # by array power percentage
+        # another alternative, calculate average POArat and TempFactor separately
+        poa_rat = [pa / pr for pa, pr in zip(poa_actual, poa_ref)]
+        tempfactor = list(map(_temp_factor, gammas, t_ref, t_actual))
+        poa_rat_x_temp_factor = adjust(  # modelchain outputs are all shifted
+            sum([p * t for p, t in zip(poa_rat, tempfactor)]) / num_arrays
+        )
```
alorenzo175, line 634: @cwhanse is this OK or would another method be preferred? Right now, gamma could be different for each array, and we are allowing uploading the weather data for each array, so we need some way to aggregate to the inverter level where Pac/Pdc are defined.

wholmgren:
> In order to convert from any temperature and irradiance input to cell temp and POA global, this just runs the full pvlib.ModelChain on the the reference and actual weather and grabs the results.

Here's a copy of the table of subcases 2A-1 -- 2A-4 in the powerpoint that @cwhanse provided (nice that something handles the formatting to make copy/paste just work!), plus my understanding of what this approach enables (2A-5 -- 2A-8)

Subcase | Time step | Reference weather | Reference DC power | Reference temperature
-- | -- | -- | -- | --
2A-1 | Hourly | POA | Yes | Cell/module
2A-2 | Hourly | POA | No | Cell/module
2A-3 | Monthly | POA | No | Cell/module
2A-4 | Hourly | GHI, DNI, DHI | No | Ambient
2A-5 | Hourly | POA | No | Ambient
2A-6 | Hourly | GHI, DNI, DHI | No | Cell/module
2A-7 | Hourly | effective irradiance | No | Ambient
2A-8 | Hourly | effective irradiance| No | Cell/module

Is that right? If so, I think that's ok. Less efficient some situations, but the flexibility win is pretty big. We could pass no-op functions for dc_model, losses_model, and ac_model to make it clear that we're not computing those quantities.

```python
def no_op(mc):
    return mc

ModelChain(dc_model=no_op,...)
```

Probably not worth the effort with our well-defined metadata requirements - we're not likely to run into a problem when ModelChain tries to calculate those quantities that we don't actually need.

Does your approach also enable the same permutations with/without reference DC power? I didn't dig that hard.

Maybe we should turn `modechain._irrad_for_celltemp` into a public function if we're going to call it directly here.

alorenzo175:
> Is that right? If so, I think that's ok. Less efficient some situations, but the flexibility win is pretty big. We could pass no-op functions for dc_model, losses_model, and ac_model to make it clear that we're not computing those quantities.

 yes


> Does your approach also enable the same permutations with/without reference DC power? I didn't dig that hard.

and yes

> Maybe we should turn `modechain._irrad_for_celltemp` into a public function if we're going to call it directly here.

sounds good to me

alorenzo175:
@lboeman do you want to take an initial look? there will likely be some minor changes coming later

lboeman:
> @lboeman do you want to take an initial look? there will likely be some minor changes coming later

Yep, I'll take a look through it right now.

lboeman (commented):
Looks very reasonable to me :+1:

api/solarperformanceinsight_api/compute.py
cwhanse: If I follow this code correctly: `poa_rat` and `tempfactor` are per array (per array derived from `poa_actual`. `poa_rat_x_temp_factor` is an average of the adjustment factor (averaged over arrays).

If I have that right, then I think this method is OK. The case where it could be improved is if a user can upload ref_pdc per array, but I don't think we envision that option. With per-array pdc, we would calculate `pdc_ref_adj` per array, sum the result then apply the inverter efficiency factor and clipping limit.

alorenzo175: Yes, you have that right, and yes, we don't plan on supporting per-array pdc.

api/requirements.txt
```diff
@@ -26,7 +26,7 @@ prometheus-fastapi-instrumentator==5.5.1
 pandas==1.1.4
 pymysql==0.10.1
 sqlalchemy==1.3.20
-pvlib==0.9.0a4
+git+https://github.com/pvlib/pvlib-python@c17f728#egg=pvlib
```
wholmgren, line 29: I'll make an a5 before this PR is merged

api/solarperformanceinsight_api/compute.py
```diff
@@ -444,3 +507,293 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
...
+        t_actual = _get_temp(
+            actual_weather, chain_actual.results.cell_temperature, inv.arrays, tshift
+        )
+
+        # mean of array POArat * TempFactor for this inverter
+        # could make more sense to use weighted mean with weights set
+        # by array power percentage
+        # another alternative, calculate average POArat and TempFactor separately
+        poa_rat = [_zero_div(pa, pr) for pa, pr in zip(poa_actual, poa_ref)]
+        tempfactor = list(map(_temp_factor, gammas, t_ref, t_actual))
```
wholmgren, line 673: Is there a reason you're using a list comprehenion and then a map? The style difference threw me off.

api/solarperformanceinsight_api/compute.py
```diff
@@ -444,3 +507,293 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
...
+    gammas = [arr._gamma for inv in inverters for arr in inv.arrays]
+    if any([g is None for g in gammas]):
+        raise TypeError(
+            "Currently unable to compare predicted and actual performance for "
+            "PVsyst specified arrays."
+        )
+
+    avg_gamma = sum(
+        [
+            sum([arr._gamma for arr in inv.arrays]) / len(inv.arrays)  # type: ignore
```
wholmgren, line 765: Does this work?

```python
from statistics import mean
avg_gamma = mean([
    mean([arr._gamma for arr in inv.arrays])
    for inv in inverters
])
```

e.g. 

```python
nested_list = [[1, 2, 3], [1, 2], [2, 3]]
mean([
    mean([a for a in b])
    for b in nested_list
])
2.0
```

wholmgren (commented):
just a handful of nits. overall looks very good.

api/requirements.txt
wholmgren: https://pypi.org/project/pvlib/0.9.0a5/

api/solarperformanceinsight_api/compute.py
```diff
@@ -444,3 +507,293 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
...
+        )
+        poa_actual = _irrad_for_celltemp(
+            chain_actual.results.total_irrad, chain_actual.results.effective_irradiance
+        )
+        # use cell temperature from the pvlib modelchain
+        # If air temp + wind speed were supplied, they are converted
+        # to cell temperature via the array's temperature model.  If
+        # module_temperature was supplied and the temperature model is
+        # sapm, it is converted to cell_temperature, otherwise module
+        # temperature is used in place of cell_temperature
```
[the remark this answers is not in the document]
alorenzo175: No. If you just give `ModelChain` module_temperature and the temperature model is not SAPM, it runs a default `air_temp = 20` and `wind_speed = 0` through the temperature model to calculate cell_temperature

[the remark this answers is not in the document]
alorenzo175: It the model is SAPM, it does appropriately convert from module to cell temp

api/solarperformanceinsight_api/compute.py
alorenzo175: not particularly. `_zero_div` came later, before it was just `[ pa / pr for ...]` so the comprehension made more sense

api/solarperformanceinsight_api/compute.py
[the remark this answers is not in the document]
wholmgren: ugh https://github.com/pvlib/pvlib-python/issues/1113

api/solarperformanceinsight_api/compute.py
alorenzo175: wow, did not know the statistics module existed...

api/solarperformanceinsight_api/models.py
```diff
@@ -392,6 +393,10 @@ def __init__(self, **data):
                 self.temperature_model_parameters._modelchain_temperature_model,
             ),
         )
+        if isinstance(self.module_parameters, PVWattsModuleParameters):
+            self._gamma = self.module_parameters.gamma_pdc
+        elif isinstance(self.module_parameters, CECModuleParameters):
+            self._gamma = self.module_parameters.gamma_r
```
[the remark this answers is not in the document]
alorenzo175: with `PVsystModuleParameters`, yes

api/solarperformanceinsight_api/compute.py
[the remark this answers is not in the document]
cwhanse: @wholmgren would this SPI issue be fixed by [pvlib/pvlib-python#1113](https://github.com/pvlib/pvlib-python/issues/1113)?

api/solarperformanceinsight_api/models.py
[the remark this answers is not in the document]
cwhanse: @wholmgren [pvlib #1190](https://github.com/pvlib/pvlib-python/pull/1190) hung up on crafting a test case.

alorenzo175:
Ah, I did forget to drop feb. 29 from the actual data if not present in predicted. At this point, I think a follow up issue/PR for that makes sense

api/solarperformanceinsight_api/compute.py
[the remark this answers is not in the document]
wholmgren: Well, it depends on what we mean by made worse. Removing the defaults would mean that the SPI code would break unless we handled the cases properly. But explicitly handling those cases would have prevented my confusion.

api/solarperformanceinsight_api/models.py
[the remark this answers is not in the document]
wholmgren: @cwhanse no worries. we'll go ahead with this and add support for pvsyst module parameters in a follow up PR after you're done.

wholmgren (approved):

PR #170, opened 2021-03-11
Usecase 2A dashboard implementation 

### LINKED ISSUE #166: Dashboard Support for Usecase 2A/2B
After the call Monday, it's clear that some data restrictions will need to move levels to support 2A and 2B. We had discussed how to make the dashboard form more simple for the compare performance use cases, but we may have oversimplified. I think the restrictions we have on the data are important in ensuring that we can proceed with the computation without any gotchas or incompatibilities that only crop up once you try and run the comparison. I think we can support 2A/B while still avoiding those issues by making a form like 
image

Then we would be able to support different weather formats for the two uploads, and different time indices. We would add some validation to make sure the predicted time index could be shifted to the actual/expected time index.

_Originally posted by @alorenzo175 in https://github.com/SolarPerformanceInsight/solarperformanceinsight/issues/65#issuecomment-768446066_

lboeman:
closes #166


lboeman:
Updates to allow users to specify jobs of usecase 2A. Should also extend to usecase 2B by enabling a radio button. 

I've updated the monthly data prompt for accuracy
> When comparing monthly data users are expected to provide monthly totals of energy and POA insolation and monthly average daytime cell temperature. Monthly data should be provided as one set for the entire system. 

and added some explanation text to the Time parameters section when predicted data is expected

>Predicted data may be provided for a year previous to the time index described below. An attempt will be made to shift predicted data by full years to match the index. The timestep and the month, day, and time of the start and end of the predicted data must match the index. Any extra timestamps in the predicted data will be ignored, for instance when February 29th exists in the predicted data but not in the index. 

Screenshot of the entire compare job specification form for usecase 2A:
Screenshot_2021-03-12 solarperformanceinsight(1)

dashboard/src/components/jobs/parameters/JobParams.vue
```diff
@@ -9,137 +9,52 @@
...
             <div v-if="isMonthly">
               <p>
-                Monthly data is expected to contain monthly averages actual
-                performance and weather for the entire system. Weather data is
-                expected to include total plane of array insolation and average
-                daytime cell temperature.
+                When comparing monthly data users are expected to provide
+                monthly totals of energy and POA insolation and monthly average
+                daytime cell temperature. Monthly data should be provided as one
+                set for the entire system.
```
alorenzo175, line 27:
```suggestion
                set for the entire system and include all 12 months.
```

dashboard/src/components/jobs/parameters/JobParams.vue
```diff
@@ -9,137 +9,52 @@
...
+                  "
+                >
+                  Predicted data may be provided for a year previous to the time
+                  index described below. An attempt will be made to shift
+                  predicted data by full years to match the index. The timestep
+                  and the month, day, and time of the start and end of the
+                  predicted data must match the index. Any extra timestamps in
+                  the predicted data will be ignored, for instance when February
+                  29th exists in the predicted data but not in the index.
+                </template>
```
alorenzo175, line 54:
```suggestion
                  Predicted data may be provided for a different year from the time
                  index described below. An attempt will be made to shift
                  predicted data by full years to match the index. The timestep
                  and the month, day, and time of the start and end of the
                  predicted data must match the index. Any extra timestamps in
                  the predicted data will be ignored, for instance when February
                  29th exists in the predicted data but not in the index. When 
                  shifting predicted data from a non-leap year to a leap year, 
                  February 29 will be dropped from the analysis.
                </template>
```

dashboard/src/components/jobs/parameters/DataParamHandler.vue
```diff
@@ -0,0 +1,159 @@
...
+    } else if (this.jobClass == "compare") {
+      if (this.jobTypeParams["compare"] == "expected and actual performance") {
+        // Single set of "data parameters" for both expected and actual performance
+        return ["expected and actual"];
+      } else if (
+        this.jobTypeParams["compare"] == "predicted and actual performance"
+      ) {
+        return ["predicted", "actual"];
+      } else {
+        return ["predicted", "expected"];
```
alorenzo175, line 125: would transposing these also swap them on the page so actual/expected is above the time parameters?

dashboard/src/components/jobs/parameters/DataParams.vue
```diff
@@ -0,0 +1,292 @@
...
+      performance_granularity: "system"
+    };
+  }
+  created() {
+    if (this.jobClass == "compare") {
+      // Ensure the correct data availability is set for compare usecase
+      // This field will not be included in the posted job JSON for
+      // dataTypes other than predicted.
+      if (this.dataType == "predicted") {
+        this.data_available = "weather only";
```
alorenzo175, line 228: can we make 'weather and AC performance' the default? Good to keep the same order, but we should direct people to that one probably

alorenzo175: for both predicted and actual, good to set irradiance to poa and temp to cell

dashboard/src/components/jobs/parameters/DataParamHandler.vue
```diff
@@ -0,0 +1,159 @@
...
+    if (this.jobClass == "calculate") {
+      if (this.jobTypeParams["calculate"] == "predicted performance") {
+        return ["predicted"];
+      } else {
+        return ["expected"];
+      }
+    } else if (this.jobClass == "compare") {
+      if (this.jobTypeParams["compare"] == "expected and actual performance") {
+        // Single set of "data parameters" for both expected and actual performance
+        return ["expected and actual"];
```
alorenzo175, line 119: should only be actual as all expected data is calculate from actual

dashboard/src/components/jobs/parameters/DataParamHandler.vue
alorenzo175, line 125:
```suggestion
        return ["predicted", "actual"];
```
will provide actual weather to compute expected performance

alorenzo175: but maybe that will add the performance questions which are irrelevant in this case. "actual weather" would be ok too, or worry about it later when it's unhidden

alorenzo175 (approved):
seems to all work, so a few style things

dashboard/src/components/jobs/parameters/DataParamHandler.vue
lboeman: With the `<slot>` header override added, we can adjust the name displayed for the user, and I agree that "actual weather" makes more sense there. I took a quick look at #171 and it looks like this job type has an `expected_data_params` property, so I think at least for the model/control portion of this code using the actual/expected/predicted terminology makes sense and if we need to adjust the front end that can be done with templates or a getter for the text to display to the user.

dashboard/src/components/jobs/parameters/DataParamHandler.vue
lboeman: I've adjusted this to be a little more clear about this values function. I needed/wanted a special flag value to represent one set of data parameters for the whole job, as are used for the `expected and actual performance` compare job. I've changed this flag value to `all_data` and added a `<slot>` around the generated header in the DataParams component, so that the header can be overriden.

PR #177, opened 2021-03-15
Drop leap day in actual data if not in predicted

### LINKED ISSUE #173: UC2A: Drop Feb. 29 if present in actual and not predicted
Follow on from #169

alorenzo175:
closes #173

Does not differentiate between uploaded reference/predicted data that simply has no data for the leap day vs reference/predicted data that was shifted to match the actual data year

### DIFF
--- a/api/solarperformanceinsight_api/compute.py
+++ b/api/solarperformanceinsight_api/compute.py
@@ -555,10 +555,10 @@ def _get_missing_leap_days(dfs: List[pd.DataFrame]) -> Set[dt.datetime]:
             if not feb29.any():
                 continue
             nans = pd.isnull(grp.loc[feb29])
-            # if not all data is nan on feb29, don't drop
-            if not nans.all().all():
-                continue
-            thisset |= set(grp.index[feb29].to_list())
+            # all data is nan on feb29, so user did not include date in
+            # upload (likely) or just specified nan values (less likely)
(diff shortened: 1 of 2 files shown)


api/solarperformanceinsight_api/compute.py
```diff
@@ -529,9 +539,36 @@ def _get_temp(
...
+
+        for _, grp in leap_df.groupby(leap_df.index.year):  # type: ignore
+            feb29 = grp.index.is_leap_year & (grp.index.dayofyear == 60)  # type: ignore
+            if not feb29.any():
+                continue
+            nans = pd.isnull(grp.loc[feb29])
+            # if not all data is nan on feb29, don't drop
+            if not nans.all().all():
+                continue
+            thisset |= set(grp.index[feb29].to_list())
```
wholmgren, line 561:
```suggestion
            # all data is nan on feb29, so user did not include date in
            # upload (likely) or just specified nan values (less likely)
            if nans.all().all():
                thisset |= set(grp.index[feb29].to_list())
```

wholmgren: This is based on my understanding that data uploaded by a user is reindexed to `pd.date_range(start, end, freq)`, so nans are inserted into the data.

wholmgren (approved):

PR #181, opened 2021-03-16
Add unit selection to results page

lboeman:
Adds another select option if the variable is a power or irradiance value as seen below. If the user selects a new unit and clicks "add to plot" again, the units are updated. 
Screenshot from 2021-03-16 14-21-39


alorenzo175 (approved):
Cool. And a selector and conversion for the table will come later?

lboeman:
> Cool. And a selector and conversion for the table will come later?

Ah, yep thanks I'll add that to this PR.

lboeman:
Added summary table unit selection as well, and updated the formatter to round to allow for precision to the nearest watt instead of always displaying an integer value.. Is that reasonable?

dashboard/src/utils/unitConversion.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+const energyUnitValues: Record<string, number> = {
+  Wh: 1,
+  kWh: 1000,
+  MWh: 1000000,
+  GWh: 1000000000
+};
+
+const insolationUnitValues: Record<string, number> = {
```
alorenzo175, line 17:
```suggestion
const irradianceUnitValues: Record<string, number> = {
```

dashboard/src/utils/unitConversion.ts
```diff
@@ -0,0 +1,53 @@
...
+  "W/m^2": 1,
+  "kW/m^2": 1000,
+  "MW/m^2": 1000000,
+  "GW/m^2": 1000000000
+};
+
```
alorenzo175, line 23:
```suggestion

const insolationUnitValues: Record<string, number> = {
  "Wh/m^2": 1,
  "kWh/m^2": 1000,
  "MWh/m^2": 1000000,
  "GWh/m^2": 1000000000
};
```

dashboard/src/utils/unitConversion.ts
```diff
@@ -0,0 +1,53 @@
...
+function getUnitValues(units: string) {
+  if (units in powerUnitValues) {
+    return powerUnitValues;
+  } else if (units in energyUnitValues) {
+    return energyUnitValues;
+  } else if (units in insolationUnitValues) {
+    return insolationUnitValues;
```
alorenzo175, line 30:
```suggestion
    return insolationUnitValues;
  } else if (units in irradianceUnitValues) {
    return irradianceUnitValues;
```

dashboard/src/components/jobs/CustomResultPlots.vue
```diff
@@ -200,14 +228,27 @@ export default class CustomPlots extends Vue {
   get timezone() {
     return this.job.definition.parameters.time_parameters.timezone;
   }
+
+  get convertableVariables() {
+    const convertable: Array<string> = [];
+    for (const index in this.selectedVariables) {
+      const variable = this.selectedVariables[index];
+      // @ts-expect-error
+      convertable[index] = ["W", "Wh", "W/m^2"].includes(
```
alorenzo175, line 237:
```suggestion
      convertable[index] = ["W", "Wh", "W/m^2", "Wh/m^2"].includes(
```

alorenzo175 (commented):
I would actually keep everything as integers. Usually when your changing units like that it's because those extra digits don't matter. Maybe @wholmgren has an opinion on that

alorenzo175:
very nice!

PR #185, opened 2021-03-17
make inf * 0 = 0 for weather adjusted performance

alorenzo175:
occurs when poa_ref = 0 so poa_factor = inf, but pac_ref is also zero,
so set the adjusted pac to 0 instead of nan

### DIFF
--- a/api/solarperformanceinsight_api/compute.py
+++ b/api/solarperformanceinsight_api/compute.py
@@ -497,14 +497,14 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
     save_results_to_db(job.object_id, result_list, si)
 
 
-def _zero_nans(out, a):
-    a_almost_zero = abs(a) < 1e-16
+def _zero_nans(out: pd.Series, a: pd.Series) -> pd.Series:
+    a_almost_zero = a.abs() < 1e-16  # type: ignore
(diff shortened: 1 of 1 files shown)


api/solarperformanceinsight_api/compute.py
```diff
@@ -497,19 +497,29 @@ def compare_expected_and_actual(job: models.StoredJob, si: storage.StorageInterf
...
+
+
 def _temp_factor(gamma, t_ref, t_actual):
     t0 = 25.0
     return _zero_div(1 - gamma * (t_actual - t0), 1 - gamma * (t_ref - t0))
 
 
+def _inf_mul(a, b):
+    """set 0 * inf = 0"""
+    out = a.mul(b, axis=0)
```
wholmgren, line 519: out of curiosity, why do you need to use `mul` here but you can use `/` above?

wholmgren (approved):

api/solarperformanceinsight_api/compute.py
alorenzo175: `/` happens to operate on pd.Series where `a` in `_inf_mul` is a dataframe and otherwise broadcasts to the wrong axis. Certainly somewhat dangerous and ambiguous; I'll change it a bit

PR #186, opened 2021-03-17
Allow no header CSV

### LINKED ISSUE #158: Allow CSVs without a header
maybe if CSV header row is 0, then don't use any header row

lboeman:
Allows users to set the header row to 0, and creates some stub headers like "Column 1, Column 2" etc.
closes #158

### DIFF
--- a/dashboard/src/components/jobs/CSVMapper.vue
+++ b/dashboard/src/components/jobs/CSVMapper.vue
@@ -2,7 +2,7 @@
 component that handles the csv headers of one file and master mapping for the
 system.
 Takes the following props:
-  - headers: Array<string> - the csv headers to be mapped.
+  - headers: Array<CSVHeader> - the csv headers to be mapped.
   - granularity: string: What part of the spec the data is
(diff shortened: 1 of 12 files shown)


lboeman:
This now allows for missing headers by setting the header line to 0. I've added a help box to assist users in utilizing this. Also fixes a bug that @wholmgren brought up, where a file without an index (time) header fails to parse/post correctly. Parsing can now allow for a single empty header, because "" is a valid object key. I've added a validation step to check CSVs for duplicate headers, as they would not raise an error but would cause unexpected behavior when posting.

@alorenzo175 Can you give this a quick poke?

alorenzo175:
Is this supposed to happen in Data Rows? 
image

lboeman:
> Is this supposed to happen in Data Rows? csv headers = 0
> image

Nope, thanks. Caused by last minute debugging junk.

wholmgren (approved):
we'll fix any bugs in production, er dev

PR #199, opened 2021-03-25
Minor bugs

### LINKED ISSUE #198: Demo bugs
- [x] No error message when lat/lon outside bounds of elevation lookup.
- [x] Module lookup  says "inverter" parameters.
- [x] Get rid of "actual" on expected vs actual job setup page.
- [x] Extra space before "AC" in the graph legend.
- [x] Disable the calculate PR button
- [x] Change "effective irradiance" button to "effective irradiance or plane of array global"

lboeman:
closes #198 
Most bugs are handled thoroughly, but the labelling for effective irradiance here is just a bandaid. I'm checking for the "effective_irradiance" variable during the data declaration/mapping process and replacing it with "Effective Irradiance or Plane of Array Global Irradiance". On the results side, I'm leaving everything as "effective irradiance". Does that sound reasonable?

### DIFF
--- a/dashboard/src/components/model/System.vue
+++ b/dashboard/src/components/model/System.vue
@@ -108,8 +108,9 @@ export default class SystemView extends ModelBase {
           "name"
         ] = `System with name "${this.parameters.name}" already exists.`;
         valid = false;
+      } else {
+        delete this.extraErrors["name"];
       }
-      delete this.extraErrors["name"];
     }
     return valid;
   }
--- a/dashboard/src/components/model/System.vue
+++ b/dashboard/src/components/model/System.vue
@@ -24,6 +24,7 @@
       field-name="longitude"
     />
     <model-field
+      @change="elevationError = null"
       :parameters="parameters"
       :errors="errors"
       :definitions="definitions"
@@ -37,6 +38,7 @@
         Look Up Elevation
       </button>
     </model-field>
+    <span class="warning-text" v-if="elevationError">{{ elevationError }}</span>
     <inverters-view :inverters="parameters.inverters" :model="model" />
   </div>
 </template>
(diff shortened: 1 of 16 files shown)


wholmgren:
> Most bugs are handled thoroughly, but the labelling for effective irradiance here is just a bandaid. I'm checking for the "effective_irradiance" variable during the data declaration/mapping process and replacing it with "Effective Irradiance or Plane of Array Global Irradiance". On the results side, I'm leaving everything as "effective irradiance". Does that sound reasonable?

Yes, I think we'd need a new `ModelChain.run_model_from_poa_global` to do it right, and that is out of scope for now. Maybe there is a better way to communicate the nuance, but I don't have a good suggestion. "Effective Irradiance (Plane of Array Global may be used instead, but reflection and spectral losses will not be accounted for)" is probably too long. What do you think @cwhanse?

dashboard/src/components/model/System.vue
```diff
@@ -76,10 +84,18 @@ export default class SystemView extends ModelBase {
...
     getElevation(this.parameters.latitude, this.parameters.longitude)
-      .then((elevation: number) => (this.parameters.elevation = elevation))
-      .catch(error => console.log(error.message));
+      .then((elevation: number) => {
+        this.parameters.elevation = elevation;
+        this.elevationError = null;
+      })
+      .catch(() => {
+        this.elevationError =
+          "Elevation information could not be found. Lookup service is only available for sites in the United States.";
```
wholmgren, line 93: does it work for Hawaii and Alaska? Territories? If not, we should say continental United States.

lboeman: It does. From the service's website: "The 1/3 arc-second dataset covers nearly all the U.S. states and territories, though Alaska has only partial coverage" (https://nationalmap.gov/epqs/) perhaps more specific wording could be used here?

wholmgren: this is good enough, thanks.

wholmgren (approved):

PR #204, opened 2021-05-12
Dashboard modeled vs reference usecase

### LINKED ISSUE #203: enable compare reference performance to modeled performance
Screen Shot 2021-05-03 at 2 01 37 PM

was added to back end in #171

lboeman:
closes #203

### DIFF
--- a/dashboard/src/components/jobs/CSVUpload.vue
+++ b/dashboard/src/components/jobs/CSVUpload.vue
@@ -495,8 +495,15 @@ ${this.granularity == "system" ? "the" : "each"} ${this.granularity}).`
           source = this.job.definition.parameters;
         }
       } else if (this.dataType.includes("actual")) {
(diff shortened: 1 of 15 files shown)


alorenzo175 (commented):
Looks good except the  modeled vs weather adjusted reference  data should be in the table instead of the monthly summary

lboeman:
> Looks good except the modeled vs weather adjusted reference data should be in the table instead of the monthly summary

Thanks, I've updated the section header to display the name of the summary style results. e.g. "Modeled vs. Weather Adjusted Reference Performance". It wasn't an issue before becasue there would only be one result with a monthly index, but this usecase provides a "monthly summary" as well as the "modeled vs weather adjusted reference" result type so leaving the "Monthly Summary" header seemed confusing.

alorenzo175 (approved):

